PR #114, opened 2021-02-26
feat: brew install 🍻

jfatta:
# Add homebrew release.
The tap is pushed here:  https://github.com/auth0/homebrew-auth0-cli.

To install

```
brew install auth0/auth0-cli/auth0
```

see: https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap

## Regarding formula "name": 

_Homebrew_ resolves which formula file to install and from which public repo should get it as: 
`auth0/homebrew-{auth0-cli}/auth0`.

Eventually, auth0-cli CI/CD could push its formula to [homebrew-core](https://github.com/Homebrew/homebrew-core) and therefore offer the default support for  _brew install auth0_ (it would be really nice). 
Keep in mind submissions to homebrew core are not covered by _goreleaser_, something like this can be used: https://github.com/mislav/bump-homebrew-formula-action

see: https://docs.brew.sh/Acceptable-Formulae
 

TODO: 
- [ ] override this repo secret `GORELEASER_GITHUB_TOKEN` with an Auth0 token to push changes to the tap repo. It is currently using a personal token.
- [ ] open https://github.com/auth0/homebrew-auth0-cli to the public, otherwise the tap is not going to work without GitHub credentials.

https://user-images.githubusercontent.com/11925502/110226966-268def00-7ed2-11eb-9347-2348cd9a9548.mov


.github/workflows/goreleaser.yml
```diff
@@ -25,4 +25,4 @@ jobs:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
```
jfatta, line 28: this is to allow goreleaser changes on external repository https://github.com/auth0/homebrew-auth0-cli

https://github.com/goreleaser/goreleaser/blob/68091726b31215473f0629d10e39a7e46fd0d82e/scripts/entrypoint.sh#L8-L11

internal/cli/completion.go
```diff
@@ -8,9 +8,11 @@ import (
 
 func completionCmd(cli *cli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "completion [bash|zsh|fish|powershell]",
+		Use:   "completion",
```
jfatta, line 11: to align with other commands but also to keep the skip-login logic simple.

.goreleaser.yml
```diff
@@ -16,21 +16,39 @@ builds:
...
+  -
+    name: auth0
+    tap:
+      owner: auth0
+      name: homebrew-auth0-cli
+    commit_author:
+      name: auth0
+      email: support@auth0.com
+    homepage: https://cli.auth0.com
+    description: Auth0 CLI
```
cyx, line 47:
```suggestion
    description: Supercharge your developer workflow.
```

Widcket (approved):

PR #122, opened 2021-03-04
CLI-27: basic integration tests.

rene00:
### Description

Initial attempt at basic integration tests.

### Testing

### Checklist


rene00:
There are no github action changes included in this PR. I'm working through understanding how we can set this up safely.

pkg/auth0-cli-config-generator/main.go
```diff
@@ -0,0 +1,206 @@
...
+	Domain      string    `json:"domain"`
+	AccessToken string    `json:"access_token,omitempty"`
+	ExpiresAt   time.Time `json:"expires_at"`
+}
+
+func isLoggedIn(filePath string) bool {
+	var c config
+	var buf []byte
+	var err error
+	if buf, err = ioutil.ReadFile(filePath); err != nil {
```
johnkerl, line 75: Go is big on the backward-compability thing, and I'm glad for it. :)

That said, in 1.16 they are suggesting people switch to `os.ReadFile`

https://www.srcbeat.com/2021/01/golang-ioutil-deprecated/

pkg/auth0-cli-config-generator/main.go
```diff
@@ -0,0 +1,206 @@
+// auth0-cli-config-generator: A command that generates a valid config file that can be used with auth0-cli.
```
cyx, line 1: Thanks for these small doc touches. Makes it feel 👌

pkg/auth0-cli-config-generator/main.go
```diff
@@ -0,0 +1,206 @@
...
+	DefaultTenant string            `json:"default_tenant"`
+	Tenants       map[string]tenant `json:"tenants"`
+}
+
+type tenant struct {
+	Name        string    `json:"name"`
+	Domain      string    `json:"domain"`
+	AccessToken string    `json:"access_token,omitempty"`
+	ExpiresAt   time.Time `json:"expires_at"`
+}
```
cyx, line 69: Note for the future: we can probably consider re-using some of the same entities we have in the CLI since they resemble each other.

pkg/auth0-cli-config-generator/main.go
cyx: General question on placement: should we put this in `./cmd/` instead? If not, curious to hear your thought process.

cyx (approved):
Just some general questions on placement, but not a real blocker. 👍

pkg/auth0-cli-config-generator/main.go
rene00: Thanks 🙇 I've replaced `ioutil.ReadFile` and `ioutil.WriteFile` with `os.ReadFile` and `os.WriteFile`.

pkg/auth0-cli-config-generator/main.go
rene00: It's only real use is to generate a config for integration testing by a CI pipeline (or a noop if a valid config exists already) so storing it outside of where it could add cognitive overhead for contributors not working on tooling for integration tests was the idea. I also wanted to ensure it didnt depend on any code within `./internal/` to keep it siloed and independent. Happy to move it to `./cmd/` and though -- just let me know.

pkg/auth0-cli-config-generator/main.go
rene00: I made a [comment](https://github.com/auth0/auth0-cli/pull/122#discussion_r598350698) on keeping it siloed but also happy to re-use entities in the CLI. I didn't want to add the coupling.

PR #146, opened 2021-03-12
Refactor string flags on apps and apis [CLI-70]

Widcket:
### Description

This command encapsulates and extracts the repetitive logic to define, register, configure and prompt for the value of string flags on `apps` and `apis` commands.

- Each flag is defined once in a struct instance that acts as a single source of truth for all the commands of a given resource
- The prompt logic is abstracted away
- The flag registration and configuration logic is abstracted away

### Testing

### Checklist


internal/cli/flags.go
```diff
@@ -0,0 +1,69 @@
...
+	DefaultValue  string
+	Help          string
+	IsRequired    bool
+}
+
+func (f *Flag) Ask(cmd *cobra.Command, value interface{}) error {
+	return ask(cmd, f, value, false)
+}
+
+func (f *Flag) AskU(cmd *cobra.Command, value interface{}) error {
```
cyx, line 23: What does U stand for?

cyx (approved):
Looks great!

internal/cli/flags.go
Widcket: It stands for "Update"

PR #154, opened 2021-03-12
test: token retriever

jfatta:
Unit test the token retriever.

also:
- add mocks for auth pkg interfaces
- update: internal/auth0/actions_mock.go, apparently we missed a `make generate` call at some point.

nice to have: check affected files on CI that are coming from `make generate`, and fail.


aherzog-auth0 (approved):
LGTM

PR #163, opened 2021-03-15
fix: app row order

jfatta:
Fix apps row representation to match table header order:

**before**
image

**after**
image


Widcket (approved):
Thanks!

PR #164, opened 2021-03-15
fix: extend select page size to the actual number of options

jfatta:
Since there's no a visual clue on the `prompter` about extra options available for "scrolling", auto-extend to the actual number of options.

**before**:

image

**after**:
image


Widcket (approved):
Thanks!

PR #166, opened 2021-03-17
Fix passing client-id as a flag on qs download

Widcket:
### Description

By using `!canPrompt(cmd)` instead of `!shouldPrompt(cmd, "client-id")`, it is possible to pass the client-id as a flag:

Screen Shot 2021-03-16 at 22 07 37

Previously it resulted in an error.

### Testing

### Checklist


jfatta (approved):
👍

cyx:
Thank you!

PR #169, opened 2021-03-19
refactor: quickstarts flags

jfatta:
Standardize flags definition and usage on the `quickstarts` command.

context: https://auth0team.atlassian.net/browse/CLI-77

### DIFF
--- a/internal/cli/quickstarts.go
+++ b/internal/cli/quickstarts.go
@@ -53,7 +53,7 @@ var (
 		Name:      "Stack",
 		LongForm:  "stack",
 		ShortForm: "s",
-		Help:      "Tech/Language of the quickstart sample to download",
+		Help:      "Tech/Language of the quickstart sample to download.",
 	}
 )
 
--- a/internal/cli/quickstarts.go
+++ b/internal/cli/quickstarts.go
@@ -42,7 +42,7 @@ var (
 	}()
 
 	clientID = Flag{
-		Name:       "ClientID",
+		Name:       "Client ID",
 		LongForm:   "client-id",
 		ShortForm:  "c",
 		Help:       "Client Id of an Auth0 application.",
--- a/internal/cli/quickstarts.go
+++ b/internal/cli/quickstarts.go
@@ -41,7 +41,7 @@ var (
 		return
 	}()
 
-	clientID = Flag{
+	qsClientID = Flag{
 		Name:       "Client ID",
 		LongForm:   "client-id",
 		ShortForm:  "c",
@@ -49,7 +49,7 @@ var (
 		IsRequired: true,
 	}
 
-	stack = Flag{
+	qsStack = Flag{
 		Name:      "Stack",
 		LongForm:  "stack",
 		ShortForm: "s",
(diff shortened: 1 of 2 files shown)


internal/cli/quickstarts.go
```diff
@@ -31,15 +40,21 @@ var (
...
-)
 
-// QuickStart app types and defaults
-const (
-	qsNative    = "native"
-	qsSpa       = "spa"
-	qsWebApp    = "webapp"
-	qsBackend   = "backend"
-	_defaultURL = "http://localhost:3000"
+	clientID = Flag{
```
Widcket, line 44: Can you please prefix it with `qs` (`qsClientID`), until we move it to its own package?

internal/cli/quickstarts.go
```diff
@@ -31,15 +40,21 @@ var (
...
+		Name:       "ClientID",
+		LongForm:   "client ID",
+		ShortForm:  "c",
+		Help:       "Client Id of an Auth0 application.",
+		IsRequired: true,
+	}
+
+	stack = Flag{
```
Widcket, line 52: Idem above.

internal/cli/quickstarts.go
```diff
@@ -31,15 +40,21 @@ var (
...
-// QuickStart app types and defaults
-const (
-	qsNative    = "native"
-	qsSpa       = "spa"
-	qsWebApp    = "webapp"
-	qsBackend   = "backend"
-	_defaultURL = "http://localhost:3000"
+	clientID = Flag{
+		Name:       "ClientID",
+		LongForm:   "client ID",
```
Widcket, line 46:
```suggestion
		LongForm:   "client-id",
```

internal/cli/quickstarts.go
```diff
@@ -31,15 +40,21 @@ var (
...
 
-// QuickStart app types and defaults
-const (
-	qsNative    = "native"
-	qsSpa       = "spa"
-	qsWebApp    = "webapp"
-	qsBackend   = "backend"
-	_defaultURL = "http://localhost:3000"
+	clientID = Flag{
+		Name:       "ClientID",
```
Widcket, line 45:
```suggestion
		Name:       "Client ID",
```
The name is used to build the prompt (eg: `Client ID:`).

internal/cli/quickstarts.go
```diff
@@ -31,15 +40,21 @@ var (
...
+		ShortForm:  "c",
+		Help:       "Client Id of an Auth0 application.",
+		IsRequired: true,
+	}
+
+	stack = Flag{
+		Name:      "Stack",
+		LongForm:  "stack",
+		ShortForm: "s",
+		Help:      "Tech/Language of the quickstart sample to download",
```
Widcket, line 56:
```suggestion
		Help:      "Tech/Language of the quickstart sample to download.",
```

internal/cli/quickstarts.go
jfatta: got it, thanks

internal/cli/quickstarts.go
```diff
@@ -31,15 +40,21 @@ var (
...
+		ShortForm:  "c",
+		Help:       "Client Id of an Auth0 application.",
+		IsRequired: true,
+	}
+
+	stack = Flag{
+		Name:      "Stack",
+		LongForm:  "stack",
+		ShortForm: "s",
+		Help:      "Tech/Language of the quickstart sample to download.",
```
Widcket, line 56: Shouldn't it be required as well? (prompt not withstanding)

jfatta: yes, good catch. I was honoring the previous setup but it doesn't make sense without the stack.

Widcket (approved):
Thanks!

Issue #130, opened 2021-03-05
updating callback urls [CLI-56]


woloski:
If I do this, I expect to just add a callback URL to this app.
`go run ./cmd/auth0 apps update ngKKhR5yy9fvIXnjRzEYdJZ7YGwjheQ8 -c http://localhost:3000`

Instead it will prompt me also for name, description, etc.

Also, if I do `go run ./cmd/auth0 apps update ngKKhR5yy9fvIXnjRzEYdJZ7YGwjheQ8` I would expect to get asked name, description, type AND callback URLs (right now it doesn't ask for that).

PR #171, opened 2021-03-19
Add rules

cyx:
## Description

This adds LIST, SHOW, CREATE, UPDATE, and DELETE operations for rules. At the moment we've left out the `Order` field altogether -- we can either consider having that in the future or completely leave re-ordering in the UI since that's a lot more intuitive.

## Demo

[asciicast](https://asciinema.org/a/SFVt6vaYPdcc0GOTTnIfrUDg4)


internal/cli/flags.go
```diff
@@ -125,6 +134,10 @@ func registerBool(cmd *cobra.Command, f *Flag, value *bool, defaultValue bool, i
 
 func shouldAsk(cmd *cobra.Command, f *Flag, isUpdate bool) bool {
 	if isUpdate {
+		if !f.IsRequired {
+			return false
+		}
+
```
cyx, line 140: @Widcket @jfatta btw this was a subtle bug -- TL;DR if it's `AskU` / `RegisterU` and not required, we should prompt for it. The assumption is most of our endpoints do `PATCH` semantics so for instance updating a rule shouldn't require the user to also update the name.

internal/cli/flags.go
```diff
@@ -15,6 +15,15 @@ type Flag struct {
...
 
+// Required clones an existing flag and assigns `IsRequired` to true. This is
+// useful when there are flags with several different use cases -- e.g. create
+// requiring the flag, but update not requiring it.
+func (f *Flag) Required() *Flag {
+	clone := *f
+	clone.IsRequired = true
+	return &clone
+}
+
```
cyx, line 26: @jfatta @Widcket While iterating on this rules feature, the difference between create requiring a name and update not -- made me first repeat the same struct.

This is an approach to DRY up that. Lemme know what you think :)

internal/cli/rules.go
```diff
@@ -0,0 +1,373 @@
...
+
+auth0 rules create --name "My Rule" --template [empty-rule]"
+		`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			prepareInteractivity(cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := ruleName.Required().Ask(cmd, &inputs.Name); err != nil {
+				return err
+			}
```
cyx, line 110: This is the bit which utilizes this new proposed feature of adding `Required()` fluent semantics to only have to define a flag once and re-define it based on usecase.

internal/cli/rules.go
```diff
@@ -0,0 +1,373 @@
...
+			if err != nil {
+				return fmt.Errorf("Unable to create rule: %w", err)
+			}
+
+			cli.renderer.RuleCreate(rule)
+			return nil
+		},
+	}
+
+	ruleName.Required().RegisterString(cmd, &inputs.Name, "")
```
cyx, line 146: Ditto ^

internal/cli/flags.go
Widcket: I think [this change](https://github.com/auth0/auth0-cli/blob/170d9e59fe1346396c2cbd30f548e9f6b592584a/internal/cli/input.go#L18) fixes it.

internal/cli/rules.go
Widcket: Should't using `AskU` on the update command handle that?

Widcket: See https://github.com/auth0/auth0-cli/blob/main/internal/cli/flags.go#L83

internal/cli/flags.go
cyx: Screen Shot 2021-03-20 at 11 00 19 AM

So when I undo this change even rebased on main I still get prompted for the Name. I think despite us passing `isRequired` in `TextInput` -- we'll still end up prompting for the Name -- which isn't the intent, correct?

internal/cli/rules.go
cyx: Gotcha, yeah I think maybe I _might_ have had a different expectation. So to step back, wanna clarify:

## Given

Name: required

## Create

Should always prompt for Name, and will fail if none is supplied.

## Update

Should _not_ prompt for Name, and will not fail if none is supplied.


-- 
Is my understanding of intent correct?

Widcket: Yup. 
On create, if not passed the command should prompt for it.
On update should only prompt if [no local flag](https://github.com/auth0/auth0-cli/issues/130#issuecomment-791683725) (any at all) is passed.
If Name wasn't a required input, it should never be prompted on update, regardless of local flags.

Widcket: We should document this somewhere.

cyx: Gotcha, based on the callback URL discussion wonder if we need a different property (e.g. `AlwaysPrompt bool`) or something similar for the case of callback URL in matias' post.

In this case, if we had `AlwaysAsk: false` in the Name field for rules, we can then skip asking for it on update. What do you think?

Widcket: Update example:
https://user-images.githubusercontent.com/5055789/111882501-b7fe6600-8994-11eb-8e66-c83602a4678f.mov

internal/cli/flags.go
Widcket: That's correct, if it is not a required input it should not prompt for it on update.

internal/cli/rules.go
Widcket: > Gotcha, based on the callback URL discussion wonder if we need a different property (e.g. `AlwaysPrompt bool`) or something similar for the case of callback URL in matias' post.
> 
> In this case, if we had `AlwaysAsk: false` in the `Name` field for rules, we can then skip asking for it on update (even if there were no local flags passed). What do you think?

Yes, adding a property for that would solve this use case. In the first PoC for the refactor I [added it](https://github.com/auth0/auth0-cli/blob/fa419aa8c80946cf0fb984c6f9dce10a995c81f0/internal/cli/flags.go#L17) and then removed because it wasn't needed yet. Now it is 🙂

internal/cli/flags.go
Widcket: Good catch!

Widcket: With this change `apps update` no longer prompts for callback URL. We should add a property for always prompting for that use case then.

Widcket (changes requested):
Add a boolean property to `Flag` to signal that it should always be prompted, and modify the prompt logic to take it into account, and the PR should be ready to go 💪🏻

Widcket (approved):

PR #172, opened 2021-03-19
refactor: logs flags and examples

jfatta:
reuse the flags helpers and added examples of the commands.

image

image

image


internal/cli/logs.go
```diff
@@ -10,6 +10,28 @@ import (
...
+
+	logsNum = Flag{
+		Name:      "Num",
+		LongForm:  "num",
+		ShortForm: "n",
+		Help:      "Number of log entries to show.",
+	}
+
+	logsNoColor = Flag{
+		Name:     "NoColor",
```
Widcket, line 29:
```suggestion
		Name:     "Disable Color",
```

internal/cli/logs.go
```diff
@@ -10,6 +10,28 @@ import (
...
+		ShortForm: "c",
+		Help:      "Client Id of an Auth0 application to filter the logs.",
+	}
+
+	logsNum = Flag{
+		Name:      "Num",
+		LongForm:  "num",
+		ShortForm: "n",
+		Help:      "Number of log entries to show.",
+	}
```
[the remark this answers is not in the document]
Widcket: We're not prompting for these (atm), but let's keep the name as if we were.

internal/cli/logs.go
Widcket: Idem above.

Widcket (changes requested):
Please use `number` as the long form of the `n` flag. The rest are non-blocking nits.

Widcket (approved):
Thanks!

PR #173, opened 2021-03-19
change spinner behavior

cyx:
As I use the CLI more, I find the spinner text that gets left behind
isn't actually useful -- it's only useful while it's happening.

After the message was completed, the user will have two states:

1. It was successful, at which point the informational text that tells
   them why they're waiting is no longer useful.
2. It failed, at which point, we'd also have separate error messages
   below the spinner text.

In summary, it's good to have the text in transient form, and clear it
out when done.

## Proposal

1. `Waiting` - used just to indicate that there's activity going on. No
   message necessary since after the wait, the outcome is obvious.

2. `Spinner` - same behavior as before. Useful for scenarios like
   deleting, since there's nothing to show after, and having the
   classic:

   `Deleting application... done`

   Text is better UX.

## Before

[asciicast](https://asciinema.org/a/Cw29ogzHyw6XFvRvwE4asaGa8)

## After

[asciicast](https://asciinema.org/a/Bh4goroRh3WLzTdw79aax2OFy)

### DIFF
--- a/internal/ansi/spinner.go
+++ b/internal/ansi/spinner.go
@@ -1,7 +1,6 @@
 package ansi
 
 import (
-	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/ansi/spinner.go
```diff
@@ -10,10 +9,7 @@ import (
 
 const (
 	spinnerTextEllipsis = "..."
-	spinnerTextDone     = "done"
-	spinnerTextFailed   = "failed"
-
-	spinnerColor = "red"
+	spinnerColor        = "red"
```
Widcket, line 12: Can we make it a color more similar to the indigo (#635DFF) that is one of the 2 primary colors of the new brand?

Widcket:
@cyx since the spinner will most likely flash quickly, should we remove the text ('Loading x')?

internal/ansi/spinner.go
cyx: Sure. Just have to check it's supported by the lib.

cyx: https://github.com/briandowns/spinner/blob/master/spinner.go#L35 these are the only colors

cyx:
> @cyx since the spinner will most likely flash quickly, should we remove the text ('Loading x')?

I'm not sure where that is defined. Do you mean the provided text?

Widcket:
> > @cyx since the spinner will most likely flash quickly, should we remove the text ('Loading x')?
> 
> I'm not sure where that is defined. Do you mean the provided text?

Yup, the provided text. If the spinner will flash and disappear, it should probably have no text.

internal/ansi/spinner.go
Widcket: Cyan, then. It's the closest.

Widcket: Screen Shot 2021-03-19 at 21 45 21

cyx:
Sure I'm OK with these proposed changes. Can do them later or feel free to make the changes

Widcket:
Go ahead whenever you are available.

Widcket:
Keep in mind that now some operations will need a 'done'  confirmation, as the provided text (along with the spinner) will no longer be there:

Screen Shot 2021-03-19 at 23 08 28

cyx:
Good point.  One option is to have a version which is backwards compatible
and another which is a transient spinner. We can probably start there,  +1?

Widcket:
+1

Widcket (approved):

PR #174, opened 2021-03-19
Show scopes in `apis` subcommands instead of count

### LINKED ISSUE #150: Apis show should show scopes [CLI-71]
for `apis show` instead of the amount of scopes, show the actual scopes

image

chrisscott:
### Description

- For `apis` subcommands that show scopes use a truncated list, based on terminal width, to show the scopes instead of the count. Use an ellipsis to denote the truncation.
- For the `apis show` subcommand, add an info to note that the full scopes can be viewed using `auth0 apis scopes list <api-id>`.

The truncation function handles conversion of the length of UTF strings to characters and truncates based on terminal width. If truncation is needed by other commands this can be refactored. 

There is an off by one bug in the truncation but it errs on the side of truncating too much so won't affect display.

The padding for the left column of the table is fixed but could be refactored to use reflection to determine the max length of the `apiView` struct and apply the pre- and post-spacing in the display to account for future additions to the struct.

### References

Fixes #150 

### Checklist


internal/display/apis.go
```diff
@@ -67,13 +69,12 @@ func (r *Renderer) ApiUpdate(api *management.ResourceServer) {
...
 
 func makeApiView(api *management.ResourceServer) *apiView {
-	scopes := len(api.Scopes)
 
 	return &apiView{
 		ID:         auth0.StringValue(api.ID),
 		Name:       auth0.StringValue(api.Name),
 		Identifier: auth0.StringValue(api.Identifier),
-		Scopes:     auth0.IntValue(&scopes),
+		Scopes:     auth0.StringValue(truncateScopes(api.Scopes)),
```
Widcket, line 77: This results in the list of scopes being too large for the table view:

Screen Shot 2021-03-19 at 21 13 02

For now, we can leave the scopes count on the table view. Apps use a different view struct to render the table view: https://github.com/auth0/auth0-cli/blob/main/internal/display/apps.go#L128

internal/cli/apis.go
```diff
@@ -142,6 +142,8 @@ auth0 apis show <id>
 			}
 
 			cli.renderer.ApiShow(api)
+			cli.renderer.Newline()
+			cli.renderer.Infof("To see the full scope list, run %s", ansi.Faint("apis scopes list <api-id>"))
```
Widcket, line 146: Please use the actual id instead of `<api-id>`.

Widcket (changes requested):
Good start, need to fix the table view and add the app id to the hint 💪🏻

internal/display/apis.go
chrisscott: Good catch! Updated.

internal/cli/apis.go
```diff
@@ -131,6 +131,8 @@ auth0 apis show <id>
 			}
 
 			cli.renderer.ApiShow(api)
+			cli.renderer.Newline()
+			cli.renderer.Infof("Scopes may be truncated. To see the full list, run %s", ansi.Faint(fmt.Sprintf("apis scopes list %s", inputs.ID)))
```
Widcket, line 135:
```suggestion
			cli.renderer.Infof("Scopes may be truncated. To see the full list, run '%s'", ansi.Faint(fmt.Sprintf("apis scopes list %s", inputs.ID)))
```
```suggestion
			cli.renderer.Infof("Scopes may be truncated. To see the full list, run %s", ansi.Faint(fmt.Sprintf("apis scopes list %s", inputs.ID)))
```

internal/display/apis.go
```diff
@@ -28,24 +31,54 @@ func (v *apiView) AsTableRow() []string {
...
+
+func (v *apiTableView) KeyValues() [][]string {
+	return [][]string{
+		{"ID", ansi.Faint(v.ID)},
+		{"NAME", v.Name},
+		{"IDENTIFIER", v.Identifier},
+		{"SCOPES", strconv.Itoa(v.Scopes)},
+	}
+}
+
```
Widcket, line 70:
```suggestion

```
This can be safely removed

internal/display/apis.go
```diff
@@ -110,3 +155,35 @@ func makeScopeView(scope *management.ResourceServerScope) *scopeView {
 		Description: auth0.StringValue(scope.Description),
 	}
 }
+
+func truncateScopes(scopes []*management.ResourceServerScope) *string {
```
Widcket, line 159: Screen Shot 2021-03-22 at 13 51 50

Scopes that fit shouldn't be truncated.
Also let's not display the hint if the scopes are not truncated.

Widcket (changes requested):
Almost there, let's not truncate the scopes that fit.
Thank you!

internal/display/apis.go
chrisscott: 🤦 good catch. Refactoring this now to bring the hint into the display logic instead of the cli logic.

Widcket (approved):
Awesome! Thank you

PR #176, opened 2021-03-19
While perusing found this subtle bug

cyx:
We should be comparing the name as well.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -366,7 +366,7 @@ func shouldPromptWhenFlagless(cmd *cobra.Command, flag string) bool {
 	isSet := false
 
 	cmd.LocalFlags().VisitAll(func(f *pflag.Flag) {
-		if f.Changed {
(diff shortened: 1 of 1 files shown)


internal/cli/cli.go
```diff
@@ -366,7 +366,7 @@ func shouldPromptWhenFlagless(cmd *cobra.Command, flag string) bool {
 	isSet := false
 
 	cmd.LocalFlags().VisitAll(func(f *pflag.Flag) {
-		if f.Changed {
+		if f.Name == flag && f.Changed {
```
Widcket, line 369: This visits all the flag and if any is set, it takes it in account for the return value. If we check for the name, aren't we only taking in account the status of that single flag?
Context: https://github.com/auth0/auth0-cli/issues/130#issuecomment-791683725

cyx: Hmm, I see. Ok we can close this totally misunderstood the intent of this 👍

PR #177, opened 2021-03-20
Refactor arguments in apps and apis commands [CLI-78]

Widcket:
### Description

This command encapsulates and extracts the repetitive logic to define and prompt for the value of arguments on `apps` and `apis` commands.

Each argument is defined once in a struct instance that acts as a single source of truth for all the commands of a given resource.

### Testing

### Checklist

### DIFF
--- a/internal/cli/input.go
+++ b/internal/cli/input.go
@@ -7,14 +7,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type CommandInput interface {
+type commandInput interface {
 	GetName() string
 	GetLabel() string
 	GetHelp() string
 	GetIsRequired() bool
 }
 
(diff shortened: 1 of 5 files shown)


internal/cli/input.go
```diff
@@ -0,0 +1,41 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/auth0/auth0-cli/internal/prompt"
+	"github.com/spf13/cobra"
+)
+
+type CommandInput interface {
```
cyx, line 10: Small nit: this can be made private since we don't expect to use it outside the cli package.

cyx (commented):
Looking great!

cyx (approved):
Merge at will when done 💪

PR #182, opened 2021-03-23
apps, apis, rules: first stab at generalizing picker options [CLI-82]

cyx:
## Description

- Refactors the rules picker to be generic
- Adds picker for apps
- Adds picker for apis

## Demo

[asciicast](https://asciinema.org/a/wN35KPdLoBIdNddhEFv1w5zpc)

### DIFF
--- a/internal/cli/apis.go
+++ b/internal/cli/apis.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/auth0/auth0-cli/internal/ansi"
@@ -113,9 +114,16 @@ auth0 apis show <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Picker(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
+					cli.renderer.Infof("There are currently no APIs.")
+					return nil
+				}
+
 			} else {
 				inputs.ID = args[0]
 			}
@@ -217,9 +225,16 @@ auth0 apis update <id> --name myapi
 			var current *management.ResourceServer
 
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Picker(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
+					cli.renderer.Infof("There are currently no APIs.")
+					return nil
+				}
+
 			} else {
 				inputs.ID = args[0]
 			}
@@ -289,9 +304,16 @@ auth0 apis delete <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Picker(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
+					cli.renderer.Infof("There are currently no APIs.")
+					return nil
+				}
+
 			} else {
 				inputs.ID = args[0]
 			}
@@ -333,9 +355,16 @@ auth0 apis scopes list <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Picker(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
+					cli.renderer.Infof("There are currently no APIs.")
+					return nil
+				}
+
 			} else {
 				inputs.ID = args[0]
 			}
@@ -367,4 +396,26 @@ func apiScopesFor(scopes []string) []*management.ResourceServerScope {
 	}
 
 	return models
+}
+
+func (c *cli) apiPickerOptions() (pickerOptions, error) {
+	list, err := c.api.ResourceServer.List()
+	if err != nil {
+		return nil, err
+	}
+
+	// NOTE: because client names are not unique, we'll just number these
+	// labels.
+	var opts pickerOptions
+	for _, r := range list.ResourceServers {
+		label := fmt.Sprintf("%s %s", r.GetName(), ansi.Faint("("+r.GetIdentifier()+")"))
+
+		opts = append(opts, pickerOption{value: r.GetID(), label: label})
+	}
+
+	if len(opts) == 0 {
+		return nil, errors.New("There are currently no applications.")
+	}
+
+	return opts, nil
 }
--- a/internal/cli/apis.go
+++ b/internal/cli/apis.go
@@ -114,7 +114,7 @@ auth0 apis show <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				err := apiID.Picker(cmd, &inputs.ID, cli.apiPickerOptions)
+				err := apiID.Pick(cmd, &inputs.ID, cli.apiPickerOptions)
 				if err != nil {
 					return err
 				}
@@ -225,7 +225,7 @@ auth0 apis update <id> --name myapi
 			var current *management.ResourceServer
 
 			if len(args) == 0 {
-				err := apiID.Picker(cmd, &inputs.ID, cli.apiPickerOptions)
+				err := apiID.Pick(cmd, &inputs.ID, cli.apiPickerOptions)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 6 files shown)


internal/cli/picker_options.go
```diff
@@ -0,0 +1,32 @@
+package cli
+
+type pickerOptions []pickerOption
```
cyx, line 3: The previous iteration using `maps` had an undesirable behavior that options get randomized on every run. Formalizing this data structure to a list to have a deterministic order -- but also having methods on the list simplifies the call sites.

internal/cli/arguments.go
```diff
@@ -31,6 +32,32 @@ func (a *Argument) Ask(cmd *cobra.Command, value interface{}) error {
...
+		var err error
+		opts, err = fn()
+		return err
+	})
+
+	if err != nil {
+		return err
+	}
+
+	// TODO(cyx): Fix this up. For now everything depends on `flag`.
```
Widcket, line 49: We'd need to ad a `selectArgument`

Widcket:
Also, we need to fix the help text in `rules update`. 
It says `auth0 rules update --id rul_d2VSaGlyaW5n --name "My Updated Rule" --enabled=false`.
Should say `auth0 rules update <rule-id> --name "My Updated Rule" --enabled=false`.

internal/cli/arguments.go
```diff
@@ -31,6 +34,42 @@ func (a *Argument) Ask(cmd *cobra.Command, value interface{}) error {
 	return askArgument(cmd, a, value)
 }
 
+type pickerOptionsFunc func() (pickerOptions, error)
+
+func (a *Argument) Picker(cmd *cobra.Command, result *string, fn pickerOptionsFunc) error {
```
Widcket, line 39: Given that the other methods of args and flags are verbs, can we change this to be `Pick`?

internal/cli/arguments.go
```diff
@@ -31,6 +34,42 @@ func (a *Argument) Ask(cmd *cobra.Command, value interface{}) error {
...
+		return err
+	}
+
+	*result = opts.getValue(val)
+	return nil
+}
+
+func selectArgument(cmd *cobra.Command, a *Argument, value interface{}, options []string, defaultValue *string) error {
+	if canPrompt(cmd) {
+		input := prompt.SelectInput("", a.GetLabel(), a.GetHelp(), options, auth0.StringValue(defaultValue), true)
```
Widcket, line 63: Can we use `_select(cmd *cobra.Command, i commandInput, value interface{}, options []string, defaultValue *string, isUpdate bool)` here? `isUpdate` would be false.

Widcket: Looking at this I realized that above in line 36 `return ask(cmd, i, value, nil, true)` should be called with `isUpdate` `false`, like this: `return ask(cmd, i, value, nil, false)`. Can you please correct that as well?

Widcket: Using `_select` there would be no need to call the error handling function, because `_select` does it.

internal/cli/rules.go
```diff
@@ -264,7 +262,7 @@ func updateRuleCmd(cli *cli) *cobra.Command {
 		Short: "Update a rule",
 		Long: `Update a rule:
 
-auth0 rules update --id  rul_d2VSaGlyaW5n --name "My Updated Rule" --enabled=false
+auth0 rules update --id <rule-id> --name "My Updated Rule" --enabled=false
```
Widcket, line 265:
```suggestion
auth0 rules update <rule-id> --name "My Updated Rule" --enabled=false
```

internal/cli/rules.go
```diff
@@ -171,10 +175,7 @@ auth0 rules show <id>
 			if len(args) > 0 {
 				inputs.ID = args[0]
 			} else {
-				// TODO(cyx): Consider making a primitive for
-				// Argument to ask using a provided func.
-				var err error
-				inputs.ID, err = promptForRuleViaDropdown(cli, cmd)
+				err := ruleID.Pick(cmd, &inputs.ID, cli.rulePickerOptions)
```
Widcket, line 178: No nil check here?

```go
if inputs.ID == "" {
	cli.renderer.Infof("There are currently no rules.")
	return nil
}
```

internal/cli/rules.go
```diff
@@ -215,18 +216,15 @@ func deleteRuleCmd(cli *cli) *cobra.Command {
...
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) > 0 {
 				inputs.ID = args[0]
 			} else {
-				// TODO(cyx): Consider making a primitive for
-				// Argument to ask using a provided func.
-				var err error
-				inputs.ID, err = promptForRuleViaDropdown(cli, cmd)
+				err := ruleID.Pick(cmd, &inputs.ID, cli.rulePickerOptions)
```
Widcket, line 227: No nil check here?

```go
if inputs.ID == "" {
	cli.renderer.Infof("There are currently no rules.")
	return nil
}
```

internal/cli/rules.go
```diff
@@ -273,10 +271,7 @@ auth0 rules update --id  rul_d2VSaGlyaW5n --name "My Updated Rule" --enabled=fal
...
```
Widcket, line 274: No nil check here?

```go
if inputs.ID == "" {
	cli.renderer.Infof("There are currently no rules.")
	return nil
}
```

Widcket (changes requested):
Looks good, only some small things to address.

internal/cli/apps.go
```diff
@@ -389,11 +404,17 @@ auth0 apps update <id> --name myapp --type [native|spa|regular|m2m]
...
 
-			// Get app id
 			if len(args) == 0 {
-				if err := appID.Ask(cmd, &inputs.ID); err != nil {
+				err := appID.Pick(cmd, &inputs.ID, cli.appPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
```
Widcket, line 413: Why the nil check if the options function already returns an error if there are no apps?

internal/cli/apps.go
```diff
@@ -209,9 +217,16 @@ auth0 apps delete <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				if err := appID.Ask(cmd, &inputs.ID); err != nil {
+				err := appID.Pick(cmd, &inputs.ID, cli.appPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
```
Widcket, line 225: Why the nil check if the options function already returns an error if there are no apps?

internal/cli/apps.go
```diff
@@ -165,9 +166,16 @@ auth0 apps show <id>
...
```
Widcket, line 174: Why the nil check if the options function already returns an error if there are no apps?

internal/cli/apis.go
```diff
@@ -333,9 +355,16 @@ auth0 apis scopes list <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Pick(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
```
Widcket, line 363: Why the nil check if the options function already returns an error if there are no apps?

internal/cli/apis.go
```diff
@@ -289,9 +304,16 @@ auth0 apis delete <id>
...
```
Widcket, line 312: Why the nil check if the options function already returns an error if there are no apps?

internal/cli/apis.go
```diff
@@ -217,9 +225,16 @@ auth0 apis update <id> --name myapi
 			var current *management.ResourceServer
 
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Pick(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
```
Widcket, line 233: Why the nil check if the options function already returns an error if there are no apps?

internal/cli/apis.go
```diff
@@ -113,9 +114,16 @@ auth0 apis show <id>
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				if err := apiID.Ask(cmd, &inputs.ID); err != nil {
+				err := apiID.Pick(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
 					return err
 				}
+
+				if inputs.ID == "" {
```
Widcket, line 122: Why the empty check if the options function already returns an error if there are no apps?

internal/cli/arguments.go
```diff
@@ -31,6 +34,42 @@ func (a *Argument) Ask(cmd *cobra.Command, value interface{}) error {
...
+
+		if err := prompt.AskOne(input, value); err != nil {
+			return handleInputError(err)
+		}
+	}
+
```
Widcket, line 69: Here, in the case of non-interactivity, we need to return an error, like in `askArgument`.

internal/cli/arguments.go
cyx: Changed. Thanks for the 👁️

internal/cli/arguments.go
```diff
@@ -31,10 +32,42 @@ func (a *Argument) Ask(cmd *cobra.Command, value interface{}) error {
...
+	if canPrompt(cmd) {
+		return _select(cmd, a, value, options, defaultValue, false)
+	}
+
+	return fmt.Errorf("Missing a required argument: %s", a.GetName())
+}
+
 func askArgument(cmd *cobra.Command, i commandInput, value interface{}) error {
 	if canPrompt(cmd) {
 		return ask(cmd, i, value, nil, true)
```
Widcket, line 69:
```suggestion
		return ask(cmd, i, value, nil, false)
```

Widcket (approved):
Awesome, thanks!

Issue #106, opened 2021-02-22
Color coding app type [CLI-38]


woloski:
can we use different colors for the type of app so it's easy to parse? only for the app type, not the full row. eg native can be cyan. etc

Widcket:
Implemented in https://github.com/auth0/auth0-cli/pull/196.

PR #196, opened 2021-03-26
feat: color code app types

aherzog-auth0:
### Description

Color codes app types when displayed in list and key-value pairs. 

Screen Shot 2021-03-26 at 2 37 09 PM

### References
https://github.com/auth0/auth0-cli/issues/106


internal/display/apps.go
```diff
@@ -305,3 +305,22 @@ func interfaceSliceToString(s []interface{}) []string {
 	}
 	return res
 }
+
+func applyColor(a string) string {
```
jfatta, line 309: do you think we can unify the _friendly_ names on a single place as constants?
It is also hardcoded here: https://github.com/auth0/auth0-cli/blob/71463818fa582c1ebdca035c2d4d922b1d958da9/internal/display/apps.go#L266-L279

jfatta (approved):
Looks good 👍 
I don't have strong opinions about the colors anyway 🕶️ 

Just a comment about constants, other than that feel free to merge.

internal/display/apps.go
aherzog-auth0: yea good call 👍

Widcket:
Thanks!

PR #197, opened 2021-03-26
fix: do not fail if keychain access failed

jfatta:
`auth0` fails while trying to access the keychain from [Windows WSL (Windows Subsystem for Linux)](https://docs.microsoft.com/en-us/windows/wsl/install-win10).
Here are more details about the issue: https://askubuntu.com/questions/815327/running-gnome-keyring-on-wsl-windows-subsystem-for-linux.

The CLI works as expected with the rest of the Windows terminals (powershell, [new windows terminal](https://github.com/microsoft/terminal), etc)

The changes on this PR handle the keychain errors gracefully and continue with the logging flow. It will not store the refresh token on the keychain therefore the login will be triggered again once the access token expired.

_before these changes_
image

_after_
image


internal/cli/login.go
```diff
@@ -63,6 +63,14 @@ func RunLogin(ctx context.Context, cli *cli, expired bool) error {
...
 	cli.renderer.Infof("Tenant: %s\n", res.Tenant)
 
+	// store the refresh token
+	secretsStore := &auth.Keyring{}
+	err = secretsStore.Set(auth.SecretsNamespace, res.Tenant, res.RefreshToken)
+	if err != nil {
+		// log the error but move on
+		cli.renderer.Warnf("Could not store the refresh token locally, please expect to login again once your access token expired: %s.", err.Error())
+	}
+
```
Widcket, line 73: Isn't it necessary to set the `secretsStore` instance on `a.Secrets`?

jfatta: not anymore, because the token is stored outside of the _authorizer_ since an error is not interrupting the whole flow.

but thanks for asking this, I missed a cleanup there in the struct definition

Widcket (approved):

PR #198, opened 2021-03-26
refactor: use argument w/ picker for quickstart id

aherzog-auth0:
### Description

Changes `quickstart download` to use an `argument` instead of `flag` which allows for "picking" client id similar to apps commands. 

Screen Shot 2021-03-26 at 3 02 41 PM
Screen Shot 2021-03-26 at 3 02 50 PM


internal/cli/quickstarts.go
```diff
@@ -93,12 +98,15 @@ func downloadQuickstart(cli *cli) *cobra.Command {
...
+			if inputs.ClientID == "" {
+				err := qsIDArg.Pick(cmd, &inputs.ClientID, cli.appPickerOptions)
+				if err != nil {
+					return err
+				}
 			}
 
-			if err := qsClientID.Ask(cmd, &inputs.ClientID, nil); err != nil {
-				return err
+			if !canPrompt(cmd) {
```
Widcket, line 108: Please move this to be the first thing in the command body

Widcket (changes requested):
Please remove the flag, and move the interactivity check to the top.

aherzog-auth0:
> Please remove the flag, and move the interactivity check to the top.

I might be wrong but without the flag it seems the user can't pass an id if they wanted to, do we want to force them into the picker?

Widcket:
That's the purpose of replacing the flag with an arg, the user can still pass the arg if they want to.

Widcket:
For that you'd have to take the value from the arg, and only prompt when no arg was passed. And add the tags validator (no more than 1 arg allowed)

Widcket:
Sorry I didn't make that clearer in the card 😕

aherzog-auth0:
No worries, I think I had that originally but then added the flag back in since we still have the `stack` flag

Widcket (approved):
Thank you!

PR #201, opened 2021-03-26
CLI-80: user-blocks commands with user_id

mhsu-auth0:
### Description

Add commands for managing user-blocks by user_id

Screen Shot 2021-03-26 at 4 51 31 PM

### References

https://auth0team.atlassian.net/browse/CLI-80
https://auth0.com/docs/api/management/v2#!/User_Blocks/delete_user_blocks

https://github.com/go-auth0/auth0/blob/master/management/user.go#L423-L449

### DIFF
--- a/internal/cli/user_blocks.go
+++ b/internal/cli/user_blocks.go
@@ -1,0 +1,110 @@
+package cli
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/auth0/auth0-cli/internal/ansi"
+	"github.com/spf13/cobra"
+	"gopkg.in/auth0.v5/management"
+)
+
+var (
+	userID = Argument{
+		Name: "user_id",
+		Help: "user_id of the user.",
+	}
+)
+
+func userBlocksCmd(cli *cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "user-blocks",
+		Short: "Manage brute-force protection user blocks.",
+	}
+
+	cmd.SetUsageTemplate(resourceUsageTemplate())
+	cmd.AddCommand(listUserBlocksByUserIdCmd(cli))
+	cmd.AddCommand(deleteUserBlocksByUserIdCmd(cli))
+	return cmd
+}
+
+func listUserBlocksByUserIdCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		user_id string
+	}
+
+	cmd := &cobra.Command{
+		Use:   "listByUserId",
+		Args:  cobra.MaximumNArgs(1),
+		Short: "List user-blocks by user_id",
+		Long: `List user-blocks by user_id:
+
+auth0 user-blocks listByUserId <user_id>
+`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			prepareInteractivity(cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				inputs.user_id = args[0]
+			} else {
+				return errors.New("user_id is required.")
+			}
+
+			var userBlocks []*management.UserBlock
+
+			err := ansi.Waiting(func() error {
+				var err error
+				userBlocks, err = cli.api.User.Blocks(inputs.user_id)
+				return err
+			})
+
+			if err != nil {
+				return fmt.Errorf("Unable to load user blocks %v, error: %w", inputs.user_id, err)
+			}
+
+			cli.renderer.UserBlocksList(userBlocks)
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func deleteUserBlocksByUserIdCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		user_id string
+	}
+
+	cmd := &cobra.Command{
+		Use:   "deleteByUserId",
+		Args:  cobra.MaximumNArgs(1),
+		Short: "Delete all user-blocks by user_id",
+		Long: `Delete all user-blocks by user_id:
+
+auth0 user-blocks deleteByUserId <user_id>
+`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			prepareInteractivity(cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				inputs.user_id = args[0]
+			} else {
+				return errors.New("user_id is required.")
+			}
+
+			err := ansi.Spinner("Deleting blocks for user...", func() error {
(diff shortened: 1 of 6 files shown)


internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+var (
+	userID = Argument{
+		Name: "user_id",
+		Help: "user_id of the user.",
+	}
+)
+
+func userBlocksCmd(cli *cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "user-blocks",
```
Widcket, line 21: Please put it under `users`

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+
+	"github.com/auth0/auth0-cli/internal/ansi"
+	"github.com/spf13/cobra"
+	"gopkg.in/auth0.v5/management"
+)
+
+var (
+	userID = Argument{
+		Name: "user_id",
```
Widcket, line 14:
```suggestion
		Name: "User ID",
```

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+		Help: "user_id of the user.",
```
Widcket, line 15:
```suggestion
		Help: "Id of the user.",
```

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+
+	cmd.SetUsageTemplate(resourceUsageTemplate())
+	cmd.AddCommand(listUserBlocksByUserIdCmd(cli))
+	cmd.AddCommand(deleteUserBlocksByUserIdCmd(cli))
+	return cmd
+}
+
+func listUserBlocksByUserIdCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		user_id string
```
Widcket, line 33: Please do not use snake_case

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:   "listByUserId",
```
Widcket, line 37: Please use `list`. Entire command should be `auth0 users blocks list`.

mhsu-auth0: we can list by `user_id` or list by `email / phonenumber / user_identifier`. It's a different API call, what would a good way to name the handle the two different input types?

internal/cli/user_blocks.go
Widcket: As a subcommand: `auth0 users blocks`

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+		Args:  cobra.MaximumNArgs(1),
+		Short: "List user-blocks by user_id",
```
Widcket, line 39: Please do not include the API2 entities as-is, instead describe what the command does in the context of a single user.

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+		Long: `List user-blocks by user_id:
```
Widcket, line 40: Same as above.

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func deleteUserBlocksByUserIdCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		user_id string
```
Widcket, line 76: Please do not use snake_case.

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:   "deleteByUserId",
```
Widcket, line 80: Please use`delete`. Entire command should be `auth0 users blocks delete`.

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+		Args:  cobra.MaximumNArgs(1),
+		Short: "Delete all user-blocks by user_id",
```
Widcket, line 82: Please do not include the API2 entities as-is, instead describe what the command does in the context of a single user.

internal/cli/user_blocks.go
```diff
@@ -0,0 +1,110 @@
...
+		Long: `Delete all user-blocks by user_id:
```
Widcket, line 83: Sma as above.

internal/display/user_blocks.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (v *userBlockView) KeyValues() [][]string {
+	return [][]string{
+		[]string{"Identifier", v.Identifier},
+		[]string{"IP", v.IP},
+	}
+}
+
+func (r *Renderer) UserBlocksList(userBlocks []*management.UserBlock) {
+	resource := "userBlocks"
```
Widcket, line 28:
```suggestion
	resource := "user blocks"
```

internal/display/user_blocks.go
```diff
@@ -0,0 +1,56 @@
...
+
+	r.Heading(resource)
+
+	if len(userBlocks) == 0 {
+		r.EmptyState(resource)
+		r.Infof("No blocks for user.")
```
Widcket, line 34: `r.EmptyState(resource)` and `r.Infof("No blocks for user.")` do the same thing, please remove the last one.

Widcket (changes requested):
Please refactor to fit in with the rest of the CLI. Please use the `apps`, `apis`, and `rules` commands as a reference.

internal/cli/user_blocks.go
Widcket: The command should accept any of these as a single argument, and then disambiguate. It's the easiest for the user, that remains unaware (as they should) of those implementation details.

Widcket: e.g. `auth0 apis show` accepts an API id or the audience.

Widcket: For the end user, there's no difference.

internal/cli/user_blocks.go
Widcket: Assume the user knows nothing about API2.

internal/cli/user_blocks.go
Widcket: Assume the user knows nothing about API2.

mhsu-auth0:
moving to branch on main repo

PR #204, opened 2021-03-27
test login: support passing in audience

### LINKED ISSUE #184: Support specifying an audience in test login [CLI-84]
in the same way we allow specifying client and connection, we should allow specifying audience

cyx:
Closes #184

## Demo

[asciicast](https://asciinema.org/a/flRGBMEZMAoFzd0WNvGGaAoEh)


internal/cli/cli.go
```diff
@@ -190,6 +190,10 @@ func (c *cli) getTenant() (tenant, error) {
 		return tenant{}, fmt.Errorf("Unable to find tenant: %s", c.tenant)
 	}
 
+	if t.Apps == nil {
+		t.Apps = map[string]app{}
+	}
```
cyx, line 195: bugfix: without this we had nil panics when trying to assign to the apps map on successful login.

Widcket: Good catch!

Widcket (approved):
Nice!

PR #205, opened 2021-03-27
apps picker: allow configurability of default app

### LINKED ISSUE #190: Keep "last client id created" in memory and offer that as the default for commands that require client ID [CLI-91]
image

Instead of having to find the client id, we can offer the last client created as the default option (just store it in a global var in memory). Anywhere where we ask for client ID we should default to that. Delighter.

cyx:
Closes #190

## Demo

[asciicast](https://asciinema.org/a/qrN8xhwE335nxAusohyJQtjuv)

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -167,7 +167,7 @@ Specify your preferred application for interaction with the auth0 CLI
 				cli.renderer.Infof("Successfully removed the default application")
 			} else {
 				cli.renderer.Infof("Successfully set the default application to %s", ansi.Faint(inputs.ID))
-				cli.renderer.Infof("%s: You might wanna try `auth0 quickstarts download`", ansi.Faint("Hint:"))
+				cli.renderer.Infof("%s You might wanna try 'auth0 quickstarts download %s'", ansi.Faint("Hint:"), inputs.ID)
 			}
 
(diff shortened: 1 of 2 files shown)


woloski:
🔥chef kiss

internal/cli/apps.go
```diff
@@ -117,10 +124,60 @@ func appsCmd(cli *cli) *cobra.Command {
...
+
+			if err := cli.setDefaultAppID(inputs.ID); err != nil {
+				return err
+			}
+
+			if inputs.ID == "" {
+				cli.renderer.Infof("Successfully removed the default application")
+			} else {
+				cli.renderer.Infof("Successfully set the default application to %s", ansi.Faint(inputs.ID))
+				cli.renderer.Infof("%s: You might wanna try `auth0 quickstarts download`", ansi.Faint("Hint:"))
```
Widcket, line 170: Can you please add the client id to the hint?

Widcket: Also the hint is showing two colons after `Hint`.

cyx: Done! Good eye, thanks 💪

Widcket (approved):
Suggested a fix for a typo, otherwise LGTM.

PR #207, opened 2021-03-27
instrumentation: add telemetry around failures

cyx:
This will start collecting errors. We can continue tweaking this, but
this will get us to the basics at least.

## In scope:
- Reporting the Go version
- Reporting the OS type
- Reproting the architecture (e.g. amd64)

## Out of scope:
- Discerning the different error types we have. For now, any errors will
  get sent -- so long as you're using a build that came from one of our future releases (previous releases wouldn't have this of course).

## For developers
- We should keep using our `go install ./cmd/auth0` flow. This would skip error reporting since we're not baking the instrumentation related information when we do this (the github action would be the one doing that).


jfatta (approved):
great!

do you know if we need to ask for some kind of user consent to _report_ errors?

cyx:
> great!
> 
> do you know if we need to ask for some kind of user consent to _report_ errors?

I think the best time to do this is during the install. If we can drop a note such as: "By using the auth0 CLI you agree to <...>" -- e.g. classic legalese language.

Similarly on the non-brew install options.

Issue #200, opened 2021-03-26
Default callback for next qs is wrong [CLI-111]


woloski:
The current is localhost:3000/callback but for nextjs should be http://locahost:3000/api/auth/callback

Widcket:
Tracking on [CLI-111](https://auth0team.atlassian.net/browse/CLI-111).

PR #208, opened 2021-03-29
Special case the Next.js callback URL on `qs download` [CLI-111]

Widcket:
### Description

- When the stack is `Next.js`, checks if the callback URLs contain `http://localhost:3000/api/auth/callback`. If not, prompts to add that URL to the callbacks while keeping `http://localhost:3000` for the rest of the URLs.
- Fixes the previous logic which would not prompt if any of the URLs were set. This would be the case for Next.js, if the user created the app with the CLI and added the default suggested callback URL (`http://localhost:3000`).
- Fixes the previous logic in the case of SPAs which would not prompt to add the allowed origin and allowed web origins URLs if the callback or logout URLs were set.
- Also hides the cursor when the spinner is active.

Screen Shot 2021-03-29 at 13 35 06

Screen Shot 2021-03-29 at 13 35 23

### References

Fixes https://github.com/auth0/auth0-cli/issues/200

### Testing

### Checklist

### DIFF
--- a/internal/ansi/spinner.go
+++ b/internal/ansi/spinner.go
@@ -36,6 +36,7 @@ func loading(initialMsg, doneMsg, failMsg string, fn func() error) error {
 		s := spinner.New(spinner.CharSets[11], 100*time.Millisecond, spinner.WithWriter(os.Stderr))
 		s.Prefix = initialMsg
 		s.FinalMSG = doneMsg
+		s.HideCursor = true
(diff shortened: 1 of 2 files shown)


cyx (approved):

jfatta (approved):
LGTM

Widcket:
Thanks for the reviews!

PR #212, opened 2021-03-30
fix: logout

jfatta:
Fix logout.
https://auth0team.atlassian.net/browse/CLI-114
if you skip playing with the selection, it was taking the default value, which is “”.

with these changes:
image


Widcket (approved):
Awesome, thanks!

PR #214, opened 2021-03-30
Do not wrap logout error twice

Widcket:
### Description

This PR returns the error from `removeTenant` as-is because it already has been formatted for display.

**Before**
`Unexpected error logging out tenant: : Unexpected error clearing tenant information: secret not found in keyring`

**After**
`Unexpected error clearing tenant information: secret not found in keyring`

Errors in general need some work.

### Testing

### Checklist


jpadilla (approved):
good catch! ✨

PR #217, opened 2021-03-31
Add default Token Endpoint Auth Method on `apps create` [CLI-115]

Widcket:
### Description

This PR adds a default value of `none` for the Token Endpoint Default Auth Method property on `apps create` when the Auth0 app a public client (SPA or Native).

Otherwise, the token exchange will fail with an `Oops... unauthorized` error.

Screen Shot 2021-03-31 at 13 14 05

Screen Shot 2021-03-31 at 13 20 12

### References

Documentation: https://auth0.com/docs/get-started/dashboard/application-settings

### Testing

### Checklist

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -383,16 +383,22 @@ auth0 apps create --name myapp --type [native|spa|regular|m2m]
 
 			// Load values into a fresh app instance
 			a := &management.Client{
-				Name:                    &inputs.Name,
-				Description:             &inputs.Description,
-				AppType:                 auth0.String(apiTypeFor(inputs.Type)),
-				Callbacks:               stringToInterfaceSlice(inputs.Callbacks),
-				AllowedOrigins:          stringToInterfaceSlice(inputs.AllowedOrigins),
-				WebOrigins:              stringToInterfaceSlice(inputs.AllowedWebOrigins),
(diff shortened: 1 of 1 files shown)


aherzog-auth0 (approved):

internal/cli/apps.go
```diff
@@ -668,6 +674,17 @@ func apiAuthMethodFor(v string) *string {
 	}
 }
 
+func apiDefaultAuthMethodFor(t string) *string {
+	switch apiTypeFor(strings.ToLower(t)) {
+	case appTypeNative:
+		fallthrough
+	case appTypeSPA:
+		return auth0.String("none")
```
cyx, line 682: Just so I understand, we want `none` for both native/spa, correct?

Widcket: Exactly, both SPA and native

internal/cli/apps.go
```diff
@@ -666,6 +672,17 @@ func apiAuthMethodFor(v string) *string {
...
```
cyx, line 680:
```suggestion
	case appTypeNative, appTypeSPA:
		return auth0.String("none")
```

cyx (approved):
Minor nit, otherwise 👍

internal/cli/apps.go
Widcket: Good catch, thanks!

PR #218, opened 2021-03-31
Always cleanup the test callback URL [CLI-116]

### LINKED ISSUE #216: Remove callback url on `test login` if there was an error [CLI-116]
We are not cleaning up properly the callback URL if there was an error (e.g.a rule returning an error).

Widcket:
### Description

When the test login fails, the callback URL added to perform the test login does not get removed from the client.
This PR will always remove it, even in case of login error.

### References

Fixes #216

### Testing

### Checklist


aherzog-auth0 (approved):

cyx (approved):
Nice!

PR #219, opened 2021-04-01
CLI-80: users blocks commands

mhsu-auth0:
Original PR: https://github.com/auth0/auth0-cli/pull/201

### Description

Add commands for managing brute force protection blocks

Screen Shot 2021-03-31 at 10 47 19 PM
Screen Shot 2021-03-31 at 10 54 50 PM
Screen Shot 2021-03-31 at 10 55 27 PM

### References

https://auth0team.atlassian.net/browse/CLI-80
https://auth0.com/docs/api/management/v2#!/User_Blocks/delete_user_blocks

https://github.com/go-auth0/auth0/blob/master/management/user.go#L423-L449


internal/cli/users.go
```diff
@@ -0,0 +1,114 @@
...
+		userID string
+	}
+
+	cmd := &cobra.Command{
+		Use:   "list",
+		Args:  cobra.MaximumNArgs(1),
+		Short: "List brute-force protection blocks for a given user",
+		Long: `List brute-force protection blocks for a given user:
+
+auth0 users blocks list <User ID>
```
Widcket, line 46:
```suggestion
auth0 users blocks list <user-id>
```
```suggestion
auth0 users blocks list <User ID>
```

internal/cli/users.go
```diff
@@ -0,0 +1,114 @@
...
+`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			prepareInteractivity(cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				inputs.userID = args[0]
```
Widcket, line 53: Can you please ask the user to input the User Id? Like:

```go
if len(args) == 0 {
    if err := userID.Ask(cmd, &inputs.userID); err != nil {
	return err
    }
} else {
    inputs.userID = args[0]
}
```

You'd have to declare the userID input outside the command (eg at the top):

```go
appID = Argument{
    Name:       "User ID",
    Help:       "Id of the user.",
    IsRequired: true,
}
```

internal/cli/users.go
```diff
@@ -0,0 +1,114 @@
...
+		Long: `Delete brute-force protection blocks for a given user:
+
+auth0 users blocks delete <User ID>
+`,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			prepareInteractivity(cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				inputs.userID = args[0]
```
Widcket, line 96: Same as above, please ask the user to input the User Id.

internal/cli/users.go
```diff
@@ -0,0 +1,114 @@
...
+	return cmd
+}
+
+func deleteUserBlocksCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		userID string
+	}
+
+	cmd := &cobra.Command{
+		Use:   "delete",
```
Widcket, line 84: This could be just `auth0 users unblock`, as it's an operation that's performed on a user directly. It's unblocking the user. `auth0 users blocks list`, on the other hand, acts on a subresource of users, the user blocks. It lists them.

Widcket: (from the PoV of a CLI user)

Widcket (dismissed):
Good, just need some adjustments 💪🏻 . Please ask the user to input the user ID, and have the delete command be `users unblock`.

internal/cli/users.go
mhsu-auth0: Screen Shot 2021-04-23 at 2 22 19 PM

internal/auth/auth.go
```diff
@@ -30,7 +30,7 @@ var requiredScopes = []string{
 	"create:clients", "delete:clients", "read:clients", "update:clients",
 	"create:resource_servers", "delete:resource_servers", "read:resource_servers", "update:resource_servers",
 	"create:rules", "delete:rules", "read:rules", "update:rules",
-	"read:client_keys", "read:logs",
+	"read:client_keys", "read:logs", "read:users", "update:users",
```
cyx, line 33: We don't need delete?

internal/cli/root.go
```diff
@@ -86,6 +86,7 @@ func Execute() {
 	// so add new commands in a place that reflect its relevance or relation with other commands:
 	rootCmd.AddCommand(loginCmd(cli))
 	rootCmd.AddCommand(tenantsCmd(cli))
+  rootCmd.AddCommand(usersCmd(cli))
```
cyx, line 89: Seems to be malformatted. Do you use `gofmt` on save or something similar @mhsu-auth0 ?

internal/cli/users.go
```diff
@@ -0,0 +1,127 @@
...
+	"gopkg.in/auth0.v5/management"
+)
+
+var (
+	userID = Flag{
+		Name:       "User ID",
+		Help:       "Id of the user.",
+		LongForm:   "user-id",
+		ShortForm:  "u",
+		IsRequired: true,
```
cyx, line 17:
```suggestion
```

Given it's an argument, don't think we need all these other fields. Correct @Widcket ?

Widcket: Yes, and it should probably be using `Argument`, as it has the appropriate logic

internal/cli/users.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+
+	"github.com/auth0/auth0-cli/internal/ansi"
+	"github.com/spf13/cobra"
+	"gopkg.in/auth0.v5/management"
+)
+
+var (
+	userID = Flag{
```
cyx, line 12:
```suggestion
	userID = Argument{
```

cyx (commented):
Just a comment on Flag vs Argument.

cyx (approved):
Thanks for all the hard work!

internal/auth/auth.go
mhsu-auth0: not yet, deleting user blocks only require `update:users`

PR #221, opened 2021-04-01
Make the audience flag required in `test token`

Widcket:
### Description

In the `test token` command, if the `--audience` flag is missing, the command will produce no tokens unless there is a default audience set in the tenant.
This PR makes that flag required, so that the command will always produce a token.

### Testing

### Checklist

### DIFF
--- a/internal/cli/test.go
+++ b/internal/cli/test.go
@@ -26,7 +26,15 @@ var (
 		Name:      "Audience",
 		LongForm:  "audience",
 		ShortForm: "a",
-		Help:      "The unique identifier of the target Audience you want to access.",
+		Help:      "The unique identifier of the target API you want to access.",
+	}
+
+	testAudienceRequired = Flag{
+		Name:       testAudience.Name,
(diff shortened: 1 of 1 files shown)


internal/cli/test.go
```diff
@@ -157,7 +165,7 @@ func testTokenCmd(cli *cli) *cobra.Command {
 		Short: "Fetch a token for the given client and API",
 		Long: `Fetch an access token for the given client.
 If --client-id is not provided, the default client "CLI Login Testing" will be used (and created if not exists).
-Additionally, you can also specify the --audience and --scope to use.`,
+Additionally, you can also specify the --audience (API Identifer) and --scope to use.`,
```
aherzog-auth0, line 168: IMO `Additionally, you can also specify` makes it sound like this isn't required.

Widcket: Good catch! Thanks

aherzog-auth0 (approved):

cyx (approved):
Thanks!

PR #223, opened 2021-04-02
refactor: use app picker for test login command

aherzog-auth0:
### Description

Refactors `test login` command to use app picker for selecting application. If no applications exist, it will do and notify the user that we are creating a temporary application to test with and will remove it after. 

Screen Shot 2021-04-02 at 10 10 16 AM

Screen Shot 2021-04-02 at 10 10 44 AM

Screen Shot 2021-04-02 at 10 08 13 AM

### References

https://auth0team.atlassian.net/browse/CLI-120

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -141,9 +141,8 @@ func useAppCmd(cli *cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "use",
 		Short: "Choose a default application",
-		Long: `auth0 apps use <client-id>
-Specify your preferred application for interaction with the Auth0 CLI
-`,
+		Long: `Specify your preferred application for interaction with the Auth0 CLI`,
+		Example: `auth0 apps use <client-id>`,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			prepareInteractivity(cmd)
 		},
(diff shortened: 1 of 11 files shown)


internal/cli/apps.go
```diff
@@ -804,7 +807,7 @@ func (c *cli) appPickerOptions() (pickerOptions, error) {
 	}
 
 	if len(opts)+len(priorityOpts) == 0 {
-		return nil, errors.New("There are currently no applications.")
+		return nil, ErrNoApps
```
Widcket, line 810: Thanks

internal/cli/test.go
```diff
@@ -130,16 +156,16 @@ auth0 test login -c <id> --connection <connection>`,
 					ansi.Faint("Hint:"), inputs.ClientID)
 
 				if err := cli.setFirstCommandRun(inputs.ClientID, commandKey); err != nil {
+					cleanupTempApplication(isTempClient, cli, inputs.ClientID)
 					return err
 				}
 			}
-
+			cleanupTempApplication(isTempClient, cli, inputs.ClientID)
```
Widcket, line 163: Shouldn't we be calling the cleanup function once with `defer`?

cyx: Yeah, probably similar to the cleanup func of the callbacks PR in #218 we want it to always run.

aherzog-auth0: Good call, thanks!

internal/cli/apps.go
```diff
@@ -12,6 +12,9 @@ import (
 	"gopkg.in/auth0.v5/management"
 )
 
+// ErrNoApps signifies no applications exist in a tenant
+var ErrNoApps = errors.New("there are currently no applications")
```
cyx, line 16: Only thing I'd say is private by default if we don't think other packages will use this error.

aherzog-auth0: Good call, spaced that everything is in cli package

internal/cli/test.go
```diff
@@ -65,29 +72,45 @@ If --client-id is not provided, the default client "CLI Login Testing" will be u
...
+						return fmt.Errorf("Unable to create an app for testing the login box: %w", err)
+					}
+					inputs.ClientID = client.GetClientID()
+					isTempClient = true
+					cli.renderer.Infof("Default test application successfully created\n")
 				}
-				inputs.ClientID = client.GetClientID()
+			} else {
+				inputs.ClientID = args[0]
 			}
```
cyx, line 110: Yeah probably want `defer cleanupTempApplication(..)` somewhere here instead

Widcket (approved):
Thanks!

PR #225, opened 2021-04-02
Do not wrap 'logged out' error

Widcket:
### Description

When running the `auth logout` command when there are no logged in tenants, the error message displayed `Unable to load tenants due to an unexpected error: Not yet configured. Try 'auth0 login'.` is not really appropriate as there are no tenants to load.
This PR changes the error message to simply `Not logged in. Try 'auth0 login'.`.

### Testing

### Checklist

### DIFF
--- a/internal/cli/logout.go
+++ b/internal/cli/logout.go
@@ -21,7 +21,7 @@ func logoutCmd(cli *cli) *cobra.Command {
 			if len(args) == 0 {
 				tens, err := cli.listTenants()
 				if err != nil {
-					return fmt.Errorf("Unable to load tenants due to an unexpected error: %w", err)
+					return err // This error is already formatted for display
 				}
 
 				if len(tens) == 0 {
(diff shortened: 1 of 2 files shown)


internal/cli/logout.go
```diff
@@ -21,7 +21,7 @@ func logoutCmd(cli *cli) *cobra.Command {
 			if len(args) == 0 {
 				tens, err := cli.listTenants()
 				if err != nil {
-					return fmt.Errorf("Unable to load tenants due to an unexpected error: %w", err)
+					return err // This error is already formatted for display
```
aherzog-auth0, line 24: Is it possible for another err other than `errUnauthenticated` to be returned? Do we need to check that it is that error?

aherzog-auth0 (approved):

internal/cli/logout.go
cyx: Good Q @aherzog-auth0 . It boils down to two other error types:

- json.Unmarshal
- os.ReadFile

If we want this to be complete, we can check `errUnauthenticated`, return it as is, otherwise do an error wrap with `Unexpected error: %w` type pattern.

What do you think @Widcket ?

PR #227, opened 2021-04-02
Solidify editor mechanics

### LINKED ISSUE #189: Cancel rule creation/update [CLI-86]
Add this hint when Editor opens:

```
Hint: once you close the editor, the rule will be saved. To cancel, CTRL+C.
```

cyx:
I've looked into what github does, so they use a mix of this approach, and
the survey approach.

To keep it simple, we're only gonna go with [the manual approach](https://github.com/cli/cli/blob/trunk/pkg/surveyext/editor_manual.go) they do until
we feel the full blown prompt approach is necessary.

## When to show the notice?

For certain editors (e.g. if you're using nano, vim, vi, etc) where it actually takes over your existing CLI, we can't really show this message as it wouldn't make sense.

For this reason, we have a list of these for now, and if their editor is in that list, it wouldn't appear at all (there's no way to cancel too -- apart from going to a different terminal, and doing a kill on the process -- which would be too much).

This is illustrated in the demo below.

## Demo

[asciicast](https://asciinema.org/a/ryZUeMmbFvyM7LDyobGwtm00h)

Closes #189

### DIFF
--- a/internal/prompt/editor.go
+++ b/internal/prompt/editor.go
@@ -1,67 +1,69 @@
 package prompt
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
-	"strings"
+	"runtime"
+
+	"github.com/kballard/go-shellquote"
 )
 
-const defaultEditor = "vim"
+const (
+	defaultEditor = "vim"
+)
 
-var defaultEditorPrompt = &editorPrompt{defaultEditor: defaultEditor}
+var (
+	bom        = []byte{0xef, 0xbb, 0xbf}
+	cliEditors = []string{"emacs", "micro", "nano", "nvim", "vi", "vim", "nvim"}
+)
+
+var defaultEditorPrompt = &editorPrompt{cmd: getDefaultEditor()}
 
 // CaptureInputViaEditor is the high level function to use in this package in
 // order to capture input from an editor.
 //
 // The arguments have been tailored for our use of strings mostly in the rest
 // of the CLI even though internally we're using []byte.
-func CaptureInputViaEditor(contents, pattern string) (result string, err error) {
-	v, err := defaultEditorPrompt.captureInput([]byte(contents), pattern)
+func CaptureInputViaEditor(contents, pattern string, infoFn func()) (result string, err error) {
+	v, err := defaultEditorPrompt.captureInput([]byte(contents), pattern, infoFn)
 	return string(v), err
 }
 
 type editorPrompt struct {
-	defaultEditor string
-}
-
-// GetPreferredEditorFromEnvironment returns the user's editor as defined by the
-// `$EDITOR` environment variable, or the `defaultEditor` if it is not set.
-func (p *editorPrompt) getPreferredEditor() string {
-	editor := os.Getenv("EDITOR")
-
-	if editor == "" {
-		return p.defaultEditor
-	}
-
-	return editor
-}
-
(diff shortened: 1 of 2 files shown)


internal/prompt/editor.go
```diff
@@ -1,68 +1,86 @@
...
 		return err
 	}
 
-	cmd := exec.Command(executable, p.resolveEditorArguments(executable, filename)...)
+	cmd := exec.Command(editorExe, args[1:]...)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
+	if !isCLIEditor && infoFn != nil {
```
Widcket, line 80: What happens if we show the message anyway? Would that cause any problems? Because if not, even if the message won't be read, we won't have to maintain a hardcoded list of editors.

internal/prompt/editor.go
```diff
@@ -1,68 +1,86 @@
...
+)
+
+const (
+	defaultEditor = "vim"
 )
 
-const defaultEditor = "vim"
+var (
+	bom        = []byte{0xef, 0xbb, 0xbf}
+	cliEditors = []string{"vi", "vim", "nano"}
```
Widcket, line 19:
```suggestion
	cliEditors = []string{"vi", "vim", "nano","micro"}
```
Add [micro](https://github.com/zyedidia/micro)

internal/prompt/editor.go
cyx: IMO it's just a tad weird if we do show it. I'm mostly "ok" with the idea of maintaining the list given it's such a short list of mainstream text based editors that people use anyway.

Here's a demo showing it https://asciinema.org/a/K7wMo0sBRWtq1e1qMtyJhAiZ2

It's likely just my OCD coming into play :)

Widcket: Ok, given that the list won't change much. We can add `emacs` and `neovim` to it, and it will probably stay unchanged for a long time.

internal/prompt/editor.go
```diff
@@ -1,68 +1,86 @@
...
+)
+
+const (
+	defaultEditor = "vim"
 )
 
-const defaultEditor = "vim"
+var (
+	bom        = []byte{0xef, 0xbb, 0xbf}
+	cliEditors = []string{"vi", "vim", "nano","micro"}
```
Widcket, line 19:
```suggestion
	cliEditors = []string{"vi", "vim", "nvim", "nano", "micro", "emacs"}
```
Add `emacs` and `nvim`

internal/prompt/editor.go
```diff
@@ -1,68 +1,86 @@
...
-func CaptureInputViaEditor(contents, pattern string) (result string, err error) {
-	v, err := defaultEditorPrompt.captureInput([]byte(contents), pattern)
+func CaptureInputViaEditor(contents, pattern string, infoFn func()) (result string, err error) {
+	v, err := defaultEditorPrompt.captureInput([]byte(contents), pattern, infoFn)
 	return string(v), err
 }
 
 type editorPrompt struct {
-	defaultEditor string
+	editorCmd string
```
Widcket, line 35: nit: move the functions that have `editorPrompt` as the receiver right below the struct

Widcket (changes requested):
Looks good, just add emacs and neovim to the editor list (that will cover the vast majority of CLI text editors for a long, long time).
Also pointed out a nit, not a blocker.

Widcket (approved):
Awesome, thank you!

PR #229, opened 2021-04-02
rules editor create/update tweaks

cyx:
## Description

This adds basic changes for the create/update path. For now we're
keeping them separate with an assertion that we want the straightforward
edit paths to bypass the survey bits -- since that becomes unnecessary
(e.g. when creating a rule, there's only way to do that -- and that's by
editing the rule).

For update, we'll utilize the surveyext bits heavily bollowed from
github to have the `?` prompt.

## Demo

[asciicast](https://asciinema.org/a/KMfak9WiCxEwtxYKlg4LwbBZ2)

### DIFF
--- a/internal/cli/rules.go
+++ b/internal/cli/rules.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/auth0/auth0-cli/internal/ansi"
 	"github.com/auth0/auth0-cli/internal/auth0"
-	"github.com/auth0/auth0-cli/internal/prompt"
 	"github.com/spf13/cobra"
 	"gopkg.in/auth0.v5/management"
 )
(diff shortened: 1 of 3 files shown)


internal/cli/rules.go
```diff
@@ -39,6 +38,14 @@ var (
 		Help:      "Enable (or disable) a rule.",
 	}
 
+	ruleScript = Flag{
+		Name:       "Script",
+		LongForm:   "script",
+		ShortForm:  "s",
+		Help:       "Script contents for the rule",
```
Widcket, line 45:
```suggestion
		Help:       "Script contents for the rule.",
```

Widcket (approved):
Awesome! Thank you

PR #234, opened 2021-04-05
fix: escape sequences on windows powershell

jfatta:
Fix escape sequences for powershell. With this custom initialization for windows the newlines and colors are displayed properly. On the other hand, emojis are still missing due to a powershell default settings limitation according to this: https://stackoverflow.com/questions/49476326/displaying-unicode-in-powershell/49481797#49481797

**before**
image

**after**
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/golang/mock v1.5.0
 	github.com/golang/snappy v0.0.3 // indirect
 	github.com/google/go-cmp v0.5.5
+	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/klauspost/compress v1.11.9 // indirect
 	github.com/klauspost/pgzip v1.2.5 // indirect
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/golang/mock v1.5.0
 	github.com/golang/snappy v0.0.3 // indirect
 	github.com/google/go-cmp v0.5.5
+	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
```
jfatta, line 16: apparently we missed a `go mod vendor` from previous PRs

internal/cli/root.go
```diff
@@ -107,6 +108,11 @@ func Execute() {
 		}
 	}()
 
+	// platform specific terminal initialization:
+	// this should run for all commands,
+	// for most of the architechtures there's no requirements:
```
Widcket, line 113:
```suggestion
	// for most of the architectures there's no requirements:
```

Widcket (approved):
Thanks!

PR #241, opened 2021-04-09
Use domains for tenant name in config.json and for display

chrisscott:
Also, fix a bug and use first selection for `tenant use` if no selection is made.

### Description

Use the domain name instead of tenant name for display and re-key config.json to use domain instead of name for the `tenants` object.

### References

https://auth0team.atlassian.net/browse/CLI-130

### Testing

- Remove your config file as this contains breaking changes
- Login
- `tenants list`
- `tenants use`
- Login to another tenant and repeat the above commands, using a different tenant
- Verify the default tenant is set correctly in your config
- Use other commands that show the tenant header and verify the domain is displayed instead of the tenant name (e.g. `apis list`
- Logout
- Verify logged out tenant is removed from the config and the default is set correctly

### Checklist

### DIFF
--- a/internal/cli/tenants.go
+++ b/internal/cli/tenants.go
@@ -34,7 +34,7 @@ func listTenantCmd(cli *cli) *cobra.Command {
 
 			tenNames := make([]string, len(tens))
 			for i, t := range tens {
-				tenNames[i] = t.Name
+				tenNames[i] = t.Domain
 			}
 
 			cli.renderer.ShowTenants(tenNames)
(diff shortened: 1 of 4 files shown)


Widcket:
@chrisscott `res, err := tr.Refresh(ctx, t.Name)` is using the tenant name [here](https://github.com/auth0/auth0-cli/blob/f66171178dcc87ac46859028bf8374e3a2e35c21/internal/cli/cli.go#L140).

Widcket:
Other than that, I've followed the test steps and it works fine 💪🏻 

Screen Shot 2021-04-09 at 16 52 03

Screen Shot 2021-04-09 at 16 53 16

Screen Shot 2021-04-09 at 16 51 41

internal/cli/tenants.go
```diff
@@ -63,10 +63,10 @@ func useTenantCmd(cli *cli) *cobra.Command {
 
 				tenNames := make([]string, len(tens))
 				for i, t := range tens {
-					tenNames[i] = t.Name
+					tenNames[i] = t.Domain
 				}
 
-				input := prompt.SelectInput("tenant", "Tenant:", "Tenant to activate", tenNames, "", true)
+				input := prompt.SelectInput("tenant", "Tenant:", "Tenant to activate", tenNames, tenNames[0], true)
```
Widcket, line 69: Thanks

Widcket (approved):
LGTM

Widcket:
Thanks for doing this 💪🏻

PR #250, opened 2021-04-16
docs: link the pkg managers on the readme

jfatta:
minor doc change to link the pkg manager (if you're on a fresh OS installation, you probably don't have homebrew or scoop)


Widcket (approved):
Good idea!

PR #251, opened 2021-04-16
feat: adds app open [id] command

as-herzog:
### Description

Adds `apps open <clientID>` command for opening the app settings page in Manage (browser).

Screen Shot 2021-04-16 at 2 14 10 PM
Screen Shot 2021-04-16 at 2 20 14 PM

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/auth0/auth0-cli/internal/ansi"
 	"github.com/auth0/auth0-cli/internal/auth0"
+	"github.com/auth0/auth0-cli/internal/open"
 	"github.com/auth0/auth0-cli/internal/prompt"
 	"github.com/spf13/cobra"
 	"gopkg.in/auth0.v5/management"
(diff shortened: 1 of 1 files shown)


internal/cli/apps.go
```diff
@@ -656,6 +659,69 @@ auth0 apps update <id> -n myapp --type [native|spa|regular|m2m]`,
...
+				cli.renderer.Warnf("Couldn't open the URL, please do it manually: %s.", manageUrl)
+			}
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func formatManageURL(cfg config, id string) string {
```
aherzog-auth0, line 703: Not very familiar with manage, is there a better way to get this?

Widcket: cc @cyx

Widcket: Also, could you please move this function (and `manageDomain`) to the `open` package? https://github.com/auth0/auth0-cli/blob/main/internal/open/open.go

cyx:
Think this looks good, and works well enough for today's use case. Not sure how it would work with private installs, but that doesn't work well altogether anyway, so maybe we can tackle all that later on.

cyx (approved):
Seems simple enough and easy enough to iterate on when we need to 👍

Widcket:
Thanks! 🎉

PR #253, opened 2021-04-18
Add `tenants open` and `apis open` [CLI-101] [CLI-133]

Widcket:
### Description

- `tenants open` that opens the tenant's settings page on the Dashboard. It takes the tenant's domain as a parameter.
- `apis open` that opens the API's settings page on the Dashboard. It takes either the API Id or audience as a parameter.

This is important as it provides a handy escape hatch, because the CLI does not support every parameter available for APIs, and does not allow to configure tenants ATM.

This PR also fixes a bug that caused the open logic to not work with PUS1 tenants, that do not have the region in the URL.

### Testing

### Checklist

### DIFF
--- a/internal/cli/apis.go
+++ b/internal/cli/apis.go
@@ -113,7 +113,7 @@ func showApiCmd(cli *cli) *cobra.Command {
 		Short: "Show an API",
 		Long:  "Show an API.",
 		Example: `auth0 apis show 
-auth0 apis show <id>`,
+auth0 apis show <id|audience>`,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			prepareInteractivity(cmd)
 		},
@@ -216,8 +216,8 @@ func updateApiCmd(cli *cli) *cobra.Command {
 		Short: "Update an API",
 		Long:  "Update an API.",
 		Example: `auth0 apis update 
-auth0 apis update <id> 
-auth0 apis update <id> --name myapi`,
+auth0 apis update <id|audience> 
+auth0 apis update <id|audience> --name myapi`,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			prepareInteractivity(cmd)
 		},
@@ -291,7 +291,7 @@ func deleteApiCmd(cli *cli) *cobra.Command {
 		Short: "Delete an API",
 		Long:  "Delete an API.",
 		Example: `auth0 apis delete 
-auth0 apis delete <id>`,
+auth0 apis delete <id|audience>`,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			prepareInteractivity(cmd)
 		},
@@ -336,7 +336,8 @@ func openApiCmd(cli *cli) *cobra.Command {
 		Args:    cobra.MaximumNArgs(1),
 		Short:   "Open API settings page in Auth0 Manage",
 		Long:    "Open API settings page in Auth0 Manage.",
-		Example: "auth0 apis open <id>",
+		Example: `auth0 apis open
+auth0 apis open <id|audience>`,
 		PreRun: func(cmd *cobra.Command, args []string) {
 			prepareInteractivity(cmd)
 		},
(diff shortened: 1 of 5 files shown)


internal/cli/apis.go
```diff
@@ -324,6 +326,52 @@ auth0 apis delete <id>`,
...
+				err := apiID.Pick(cmd, &inputs.ID, cli.apiPickerOptions)
+				if err != nil {
+					return err
+				}
+			} else {
+				inputs.ID = args[0]
+			}
+
+			// Heuristics to determine if this a valid ID, or an audience value
+			if _, err := url.ParseRequestURI(inputs.ID); err == nil || len(inputs.ID) != 24 {
```
jfatta, line 354: what's `24`coming from ? is inputs.ID a url? in that case I wonder if we can make it more explicit on the command help

internal/cli/apis.go
```diff
@@ -324,6 +326,52 @@ auth0 apis delete <id>`,
...
+				if err := ansi.Waiting(func() error {
+					api, err := cli.api.ResourceServer.Read(url.PathEscape(inputs.ID))
+					if err != nil {
+						return err
+					}
+					inputs.ID = auth0.StringValue(api.ID)
+					return nil
+				}); err != nil {
+					return fmt.Errorf("An unexpected error occurred while trying to get the API Id for '%s': %w", inputs.ID, err)
```
jfatta, line 363: NAB: I know we're doing this in other places already, but I'd suggest starting error messages with lowercase (as well as avoiding ending punctuation). More context here: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

internal/cli/apps.go
```diff
@@ -669,7 +667,7 @@ func openAppCmd(cli *cli) *cobra.Command {
 		Args:    cobra.MaximumNArgs(1),
 		Short:   "Open application settings page in Auth0 Manage",
 		Long:    "Open application settings page in Auth0 Manage.",
-		Example: `auth0 apps open <id>`,
+		Example: "auth0 apps open <id>",
```
jfatta, line 670: from prev comment, If id could adopt different forms (I'm not sure) I'd suggest including some placeholders here in the example section.

internal/cli/tenants.go
```diff
@@ -93,3 +101,67 @@ func useTenantCmd(cli *cli) *cobra.Command {
...
+		},
+	}
+
+	return cmd
+}
+
+func (c *cli) tenantPickerOptions() (pickerOptions, error) {
+	tens, err := c.listTenants()
+	if err != nil {
+		return nil, fmt.Errorf("Unable to load tenants due to an unexpected error: %w", err)
```
jfatta, line 144: ditto

internal/cli/tenants.go
```diff
@@ -93,3 +101,67 @@ func useTenantCmd(cli *cli) *cobra.Command {
...
+	}
+
+	var (
+		priorityOpts, opts pickerOptions
+	)
```
jfatta, line 149:
```suggestion
          var priorityOpts, opts pickerOptions
```

jfatta (approved):
looks good. minor suggestions inline.
thanks

internal/cli/apis.go
Widcket: 24 is the string length of the ID. Audiences are usually URLs, but not necessarily, so there is that additional heuristic to help disambiguate. I'll make it more explicit on the command help and add a comment explaining the values, thanks!

internal/cli/tenants.go
Widcket: Same as above.

internal/cli/apps.go
Widcket: I went with `auth0 apps open <id|audience>` in https://github.com/auth0/auth0-cli/pull/253/commits/62c64d535f533809886487d159ef00c33b04881b

internal/cli/apis.go
Widcket: Done in https://github.com/auth0/auth0-cli/pull/253/commits/62c64d535f533809886487d159ef00c33b04881b

PR #254, opened 2021-04-19
Add `quickstarts list` [CLI-90]

Widcket:
### Description

This PR adds the `quickstarts list` command that lists the available Quickstarts:

Screen Shot 2021-04-19 at 15 45 21

### Testing

### Checklist

### DIFF
--- a/internal/display/quickstarts.go
+++ b/internal/display/quickstarts.go
@@ -1,0 +1,62 @@
+package display
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/auth0/auth0-cli/internal/auth0"
+)
+
+type quickstartView struct {
+	Stack   string
+	AppType string
+	URL     string
+}
+
+func (v *quickstartView) AsTableHeader() []string {
+	return []string{"Quickstart", "Type", "URL"}
+}
+
+func (v *quickstartView) AsTableRow() []string {
+	return []string{v.Stack, applyColor(v.AppType), v.URL}
+}
+
+func (r *Renderer) QuickstartList(qs map[string][]auth0.Quickstart) {
+	r.Heading()
+
+	var results []View
+	keys := make([]string, 0, len(qs))
+
+	for key, _ := range qs {
+		keys = append(keys, key)
+	}
+
+	sort.Strings(keys)
+
+	for _, key := range keys {
+		for _, item := range qs[key] {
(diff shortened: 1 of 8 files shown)


internal/display/quickstarts.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	sort.Strings(keys)
+
+	for _, key := range keys {
+		for _, item := range qs[key] {
+			results = append(results, &quickstartView{
+					Stack: item.Name,
+					AppType: applyColor(qsAppTypeFor(key)),
+					URL: fmt.Sprintf("https://auth0.com/docs/quickstart/%s/%s", key, item.Path),
```
jfatta, line 41: nit: it can be a const at the top to be (slightly) easier to change.

internal/display/quickstarts.go
```diff
@@ -0,0 +1,62 @@
...
+				})
+		}
+	}
+
+	r.Results(results)
+}
+
+func qsAppTypeFor(s string) string {
+	switch s {
+	case "native":
```
jfatta, line 51: not a blocker: we're starting to have this duplicated. Maybe we can define some enum _-ish_ types somewhere and reuse
https://github.com/auth0/auth0-cli/blob/ba8e0d5488be489a319bdea29093aa77c1f9ca33/internal/cli/apps.go#L57-L61

internal/display/tenants.go
```diff
@@ -1,7 +1,7 @@
 package display
 
 type tenantView struct {
-	Name 		string
+	Name string
```
jfatta, line 4: 🔥

jfatta (approved):
looks great. I left a couple of suggestions. 
thanks!

internal/display/quickstarts.go
Widcket: Yes, for the time being I've reused the constants set in `display/apps.go` in https://github.com/auth0/auth0-cli/pull/254/commits/1a95a65d85fd00ebc86ce64d201949571d954585

internal/display/quickstarts.go
Widcket: Good catch, changed in https://github.com/auth0/auth0-cli/pull/254/commits/1a95a65d85fd00ebc86ce64d201949571d954585

PR #259, opened 2021-04-23
fix: polling faster error #252

### LINKED ISSUE #252: login error: You are polling faster than the specified interval of 5 seconds
On a MacOS X environment, I installed the CLI with brew and, right after logging in, I got the following error:

```bash
 ▸    ✪ Welcome to the Auth0 CLI 🎊.
 ▸    To set it up, you will need to sign in to your Auth0 account and authorize the CLI to access the API.
 ▸    If you don't have an account, please go to https://auth0.com/signup, otherwise continue in the browser.


 ▸    Your pairing code is: MJKV-RJLB

 ▸    This pairing code verifies your authentication with Auth0.
 ▸    Press Enter to open the browser (^C to quit)


===  error

 ▸    login error: You are polling faster than the specified interval of 5 seconds.
```

Just to confirm that the login flow indeed has failed, I ran `auth0 apps list`. As expected, I got:

```
===  error

 ▸    Not logged in. Try 'auth0 login'.
```

Running the flow again (`auth0 login`) succeeded. So, feel free to close the issue if you can't reproduce, but I leave it here as a note in case you want to double check.

jfatta:
Adding a threshold to prevent this error:
```
login error: You are polling faster than the specified interval of (...)
```
fixes #252

image

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -14,10 +14,11 @@ import (
 )
 
 const (
-	clientID           = "2iZo3Uczt5LFHacKdM0zzgUO2eG2uDjT"
-	deviceCodeEndpoint = "https://auth0.auth0.com/oauth/device/code"
-	oauthTokenEndpoint = "https://auth0.auth0.com/oauth/token"
-	audiencePath       = "/api/v2/"
+	clientID               = "2iZo3Uczt5LFHacKdM0zzgUO2eG2uDjT"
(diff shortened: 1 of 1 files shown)


Widcket (commented):
If we set the leeway to 2 seconds, does it keep occurring?

jfatta:
> If we set the leeway to 2 seconds, does it keep occurring?

it is tricky because I couldn't replicate this.

Widcket (approved):
Ok, let's err on the safer side.

PR #261, opened 2021-04-23
feat: login when scopes changed

jfatta:
Check if the required scopes changed and re-triggers the login flow in that case.

The scopes used are persisted on the config (tenant entity) just like the expiration date. 

**first time running this version will trigger a login since the config files are outdated**

_running any command with scopes out of date_
image

_scopes persisted_
image

_after that, until scopes change again, the login is not required_
image


Widcket (approved):
Thanks!

PR #262, opened 2021-04-23
fix: wrong jwtconfig lifetime type

jfatta:
Exclude the `jwt_configuration.lifetime_in_seconds` file from the Clients.List operation.
We experienced the following error in some cases:

```
 ▸    json: cannot unmarshal string into Go struct field ClientJWTConfiguration.clients.jwt_configuration.lifetime_in_seconds of type int
exit status 1
```

From the investigation, it seems that legacy application are returning a `jwt_configuration.lifetime_in_seconds` of type **string** instead of **int*. Since this field is not used on the CLI, we can safely ignore it for now.

_verifying that the field is being excluded_
image

_apps list command working as expected since the excluded field is not used_
image

_replicated the issue when logged with a github account, and listing a pretty old application_
```
"jwt_configuration":{
            "lifetime_in_seconds":"36000" <----- this should be a number instead of a string.
         },
```

### DIFF
--- a/internal/cli/apps_test.go
+++ b/internal/cli/apps_test.go
@@ -11,7 +11,7 @@ import (
 	"gopkg.in/auth0.v5/management"
 )
 
-func TestClientsListCmd(t *testing.T) {
+func TestAppsListCmd(t *testing.T) {
 	// Step 1: Setup our client mock for this test. We only care about
 	// Clients so no need to bootstrap other bits.
 	ctrl := gomock.NewController(t)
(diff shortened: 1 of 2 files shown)


internal/cli/apps_test.go
```diff
@@ -11,15 +11,15 @@ import (
...
+func TestAppsListCmd(t *testing.T) {
 	// Step 1: Setup our client mock for this test. We only care about
 	// Clients so no need to bootstrap other bits.
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
 	clientAPI := auth0.NewMockClientAPI(ctrl)
 	clientAPI.EXPECT().
-		List().
+		List(gomock.Any()).
```
jfatta, line 22: creating a matcher for this could be really tricky because it is a function that takes a requests and `apply` the changes. I don't it will add a fundamental value to the test, but let me know

Widcket (approved):
Thank you!

Widcket:
From Proxyman (key not found in res body):
Screen Shot 2021-04-23 at 16 20 53

PR #263, opened 2021-04-23
CLI-140: integration tests on ci

as-herzog:
### Description

Runs integration tests in CI but skips them for forked PRs:
Screen Shot 2021-05-07 at 2 31 41 PM

### Forked ex:

Screen Shot 2021-05-07 at 2 28 55 PM

### References
https://auth0team.atlassian.net/browse/CLI-140


.github/workflows/go.yml
```diff
@@ -22,3 +22,15 @@ jobs:
...
+    - name: integration
+      # skip running this action if the PR is coming from a fork:
+      if: github.event.pull_request.head.repo.full_name == github.repository
+      run: PATH=$(go env GOPATH)/bin:$PATH make integration
+      env:
+          AUTH0_CLI_CLIENT_NAME: ${{ secrets.AUTH0_CLI_CLIENT_NAME }}
+          AUTH0_CLI_CLIENT_DOMAIN: ${{ secrets.AUTH0_CLI_CLIENT_DOMAIN }}
+          AUTH0_CLI_CLIENT_ID: ${{ secrets.AUTH0_CLI_CLIENT_ID }}
+          AUTH0_CLI_CLIENT_SECRET: ${{ secrets.AUTH0_CLI_CLIENT_SECRET }}
+          AUTH0_CLI_REUSE_CONFIG: ${{ secrets.AUTH0_CLI_REUSE_CONFIG }}
```
jfatta, line 35: I'm not sure if we need `AUTH0_CLI_REUSE_CONFIG` on the secrets. It should be always false on CI if understood the code correctly; there's no config file to reuse here.

Makefile
```diff
@@ -38,11 +38,7 @@ lint:
 
 # Build for the native platform
 build:
-	go build -ldflags "$(CTIMEVAR)" -o auth0 cmd/auth0/main.go
```
jfatta, line 41: good catch. bad resolved merge conflict I guess 🤦🏼

.github/workflows/go.yml
jfatta: what if we tweak the CLI codebase to look for all those ENV vars and skip the config file completely if the vars are present?

jfatta: nvm, we don't want to support client id/secret auth mechanism on the code base. In any case, it could look for an ENV var with the access token and skip config file, but that has nothing to do with this PR.

pkg/auth0-cli-config-generator/main.go
```diff
@@ -168,7 +180,13 @@ func main() {
 				return err
 			}
 
-			t := tenant{p.clientName, p.clientDomain, token.AccessToken, token.Expiry}
+			t := tenant{
+				Name:        p.clientName,
+				Domain:      p.clientDomain,
+				AccessToken: token.AccessToken,
+				ExpiresAt:   token.Expiry,
+				Scopes:      requiredScopes,
```
jfatta, line 188: the CLI is triggering a login flow if the scopes on the tenant are not equal to the defined on the auth package, this is a way we have to "revalidate" tokens if we change the scopes used.

Evidently, it would be hard to maintain (keep in sync) the config generator tenant and login details. I would like to explore that alternative of skip all file-related config if an access token is defined. That way the generator would be repurposed on a command to login and write the access token to a var, no need to know the tenant entity at all.

Makefile
```diff
@@ -70,13 +66,13 @@ $(GOBIN)/auth0-cli-config-generator:
...
+	auth0-cli-config-generator && commander test commander.yaml --verbose
 .PHONY: run-integration
 
 # Delete all test apps created during integration testing
 integration-cleanup:
 	./integration/test-cleanup.sh
 .PHONY: integration-cleanup
 
-integration: $(GOBIN)/auth0-cli-config-generator $(GOBIN)/commander run-integration integration-cleanup
+integration: build $(GOBIN)/auth0-cli-config-generator $(GOBIN)/commander run-integration integration-cleanup
```
jfatta, line 77: @as-herzog I kept the `integration-cleanup` when I solved merge conflicts. But I think that is swallowing potential error reports from `commander` when tests failed.

Widcket (approved):
Awesome, thank you for doing this!

PR #264, opened 2021-04-26
Make the `--no-color` flag global [CLI-143]

Widcket:
### Description

This PR makes the `--no-color` flag global, as it was only previously defined for `logs` commands. The display logic was updated to take that value into account, and direct uses of the underlying `aurora` library have been fixed to use the wrapper `ansi` package instead.

Screen Shot 2021-04-26 at 16 02 43
Screen Shot 2021-04-26 at 16 05 31

### Testing

### Checklist

### DIFF
--- a/internal/ansi/ansi.go
+++ b/internal/ansi/ansi.go
@@ -88,6 +88,12 @@ func Red(text string) string {
 	return color.Sprintf(color.Red(text))
 }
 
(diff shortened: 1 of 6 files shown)


jfatta (approved):
LGTM

cyx (approved):
Nice!

PR #268, opened 2021-04-28
Update messaging for login command

chrisscott:
### Description

- Be more clear on the next step
- Use `Device Confirmation` instead of `pairing` to match language on the Universal Login page
- Use a spinner while waiting for login to complete in the browser
- Be more concise

**Before**
login-before

**After**
login

### References

CLI-135

### Testing

Test login command to completion and also canceling it out.

### Checklist

- [x] The correct base branch is being used, if not `main`


Widcket (approved):
Awesome, thank you! Nicely done 👍🏼

PR #271, opened 2021-04-30
Fix cursor escape seqs being written to StdOut [CLI-154]

Widcket:
### Description

The spinner library writes the escape sequences to StdOut, so when piping a json output to a file it will contain invalid characters.That bug was fixed on a PR from December 2020 (https://github.com/briandowns/spinner/pull/109), however it remains unreleased.
This PR updates the spinner library to the latest commit on master, to get the fix, and configures the spinner to use StdErr.

### Testing

### Checklist


app.json
```diff
@@ -0,0 +1,51 @@
+{
```
jfatta, line 1: I think we need to remove this 😅

jfatta (approved):
LGTM

app.json
Widcket: Done.

cyx (approved):
Great find!

Issue #274, opened 2021-04-30
Update SDK to support both int/string lifetime_in_seconds [CLI-155]


cyx:
Currently lifetime_in_seconds is already int. There are older tenants though in `string`.

We'll need to fix this on the SDK level.

Widcket:
Tracking in card CLI-155.

Widcket:
SDK PR: https://github.com/go-auth0/auth0/pull/213

PR #275, opened 2021-04-30
fix: just make client.List / Read work for all cases

cyx:
This should be fixed on the SDK level. Tracked in https://github.com/auth0/auth0-cli/issues/274


Widcket (approved):
Thanks!

PR #278, opened 2021-05-01
Integration tests for auth0 apps commands and flags [CLI-95]

morganelle:
### Description
Use Commander to add integration tests for the following commands, with tests for every flag and flag value.

- `apps create`
- `apps update`
- `apps show`

Added two small helper scripts to store an app id for tests that require an existing app as well as a cleanup script to delete the fixtures created during testing.

Also fixed two small typos found during development.

### Testing

These updates can be tested by running `make integration`. Verified that the apps not prefixed with `integration-test-` were not deleted during test cleanup.

### Checklist


integration/test-cleanup.sh
```diff
@@ -0,0 +1,18 @@
...
+
+for app in $( echo "${apps}" | jq -r '.[] | @base64' ); do
+    _jq() {
+     echo "${app}" | base64 --decode | jq -r "${1}"
+    }
+
+    clientid=$(_jq '.ClientID')
+    name=$(_jq '.Name')
+
+    if [[ $name != "Default App" ]]
```
morganelle, line 13: TODO: I'm going to create the test apps with a common prefix and update this logic so only integration test apps will be deleted.

commander.yaml
```diff
@@ -19,3 +19,235 @@ tests:
...
+        app_type: native
+
+  apps create type native and check output:
+    command: auth0 apps create --name integration-test-nativeapp1 --type native --description NativeApp1
+    exit-code: 0
+    stdout:
+      contains:
+        - NAME                 integration-test-nativeapp1
+        - DESCRIPTION          NativeApp1
+        - TYPE                 Native
```
rene00, line 40: This test is failing for me due to `N` being uppercase in `contains` but lower case in the result:
```
✗ [local] 'apps create type native and check output', on property 'Stdout'

Expected
[...]
  TYPE                 native
```

```suggestion
        - TYPE                 native
```

I can see that we're already testing this command [here](https://github.com/auth0/auth0-cli/pull/278/files#diff-968aa84e08a507b385aee72591b5751aafdaac0d93109337e03ba0ba860237a5R24) but this command is the plain format. Have you considered removing this plain format test all together?

rene00: ✋  Ignore this. Error was on my side. I was running an older version of `auth0`.

Makefile
```diff
@@ -71,4 +71,10 @@ $(GOBIN)/auth0-cli-config-generator:
 
 integration: $(GOBIN)/auth0-cli-config-generator $(GOBIN)/commander
 	auth0-cli-config-generator && commander test commander.yaml
+	$(MAKE) integration-cleanup
```
rene00, line 74: perhaps a nitpick here, but we could replace recursive use of `make` with:

```
run-integration:
    auth0-cli-config-generator && commander test commander.yaml
.PHONY: run-integration

integration: $(GOBIN)/auth0-cli-config-generator $(GOBIN)/commander run-integration integration-cleanup
.PHONY: integration
```

rene00 (approved):
This an awesome PR and a great step forward for auth0-cli. Nice work 💯

rene00:
Perhaps an issue on my end but when running `make integration` I was prompted for device flow auth when running:
```
export AUTH0_CLI_CLIENT_NAME="integration" \
    AUTH0_CLI_CLIENT_DOMAIN="REDACTED" \
    AUTH0_CLI_CLIENT_ID="REDACTED" \
    AUTH0_CLI_CLIENT_SECRET="REDACTED" \
    AUTH0_CLI_REUSE_CONFIG="false" \
    AUTH0_CLI_OVERWRITE="true"
make integration
```

commander.yaml
morganelle: Removing the plain format tests that duplicate json format tests is a good question, though. I included it in case we want an indication when the formatting changes unexpectedly but it could be fussy to maintain.

morganelle:
> Perhaps an issue on my end but when running `make integration` I was prompted for device flow auth:
> 
> ```
> export AUTH0_CLI_CLIENT_NAME="integration" \
>     AUTH0_CLI_CLIENT_DOMAIN="REDACTED" \
>     AUTH0_CLI_CLIENT_ID="REDACTED" \
>     AUTH0_CLI_CLIENT_SECRET="REDACTED" \
>     AUTH0_CLI_REUSE_CONFIG="false" \
>     AUTH0_CLI_OVERWRITE="true"
> make integration
> ```

Was that before commander started executing the tests or during?

Makefile
morganelle: Updated

Widcket (approved):
Excellent, thanks for going the extra mile with this one.

integration/test-cleanup.sh
```diff
@@ -0,0 +1,18 @@
...
+
+for app in $( echo "${apps}" | jq -r '.[] | @base64' ); do
+    _jq() {
+     echo "${app}" | base64 --decode | jq -r "${1}"
+    }
+
+    clientid=$(_jq '.ClientID')
+    name=$(_jq '.Name')
+
+    if [[ $name = integration-test-* ]]
```
jfatta, line 13: I'm wondering if this would only remove the apps created during the same tests execution, otherwise it might cause issues with other CI running at the same time. 
thoughts?

morganelle: @jfatta if we have multiple integration tests on this tenant running at once, we'll want something smarter that deletes by client id. When we get CI running we'll also want to split out the test and cleanup steps so that the cleanup step is sure to run even if the tests fail.

PR #286, opened 2021-05-08
Add APIs integration tests [CLI-96]

Widcket:
### Description

This PR adds integration tests for APIs and fixes two issues where the value of the `scopes` flag would be ignored.

### Testing

### Checklist

### DIFF
--- a/integration/test-cleanup.sh
+++ b/integration/test-cleanup.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-apps=$( auth0 apps list --format json --no-input)
+apps=$( auth0 apps list --format json --no-input )
 
 for app in $( echo "${apps}" | jq -r '.[] | @base64' ); do
     _jq() {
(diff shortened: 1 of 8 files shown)


integration/test-cleanup.sh
```diff
@@ -11,9 +11,27 @@ for app in $( echo "${apps}" | jq -r '.[] | @base64' ); do
...
+    id=$(_jq '.ID')
+    name=$(_jq '.Name')
+		# TODO(jfatta): should remove only those 
+		# created during the same test session
+    if [[ $name = integration-test-api-* ]]
+    then
+        echo deleting "$name"
+        $( auth0 apis delete "$id")
+    fi
+done
```
Widcket, line 37: This can be parallelized in batches of 5 with
```sh
auth0 apis list --format json --no-input \
    | jq -r '.[] | select(.Name | startswith("integration-test-api-")) | "\(.Name) \(.ID)\n"' \
    | xargs -P 5 -n 2 bash -c 'echo Deleting $0 && auth0 apis delete $1 --force' 
```

PR #290, opened 2021-05-14
fix: use updated token once after the renewal

jfatta:
Fix bug regarding old access token being used after the renewal process due to expiration or scopes changed.

_before_
image

_after_
image


cyx (approved):
Nice! 👍

Widcket (approved):

Widcket:
Thanks!

PR #291, opened 2021-05-14
feat: config init command

as-herzog:
### Description

Adds `auth0 config init` command to initialize cli config from environment variables (non-interactive flow). This is a replacement for the `cli-config-generator` used in integration tests. 

Screen Shot 2021-05-17 at 3 15 16 PM


pkg/auth0-cli-config-generator/main.go
```diff
@@ -12,8 +12,9 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
-	"time"
 
+	"github.com/auth0/auth0-cli/internal/auth"
```
jfatta, line 16: can we export the configuration bits more formally? If the config-generator wouldn't need to import /internal/. 
what's the minimum structs or functions that are needed from the CLI config to be initialized externally?

jfatta: what if we actually provide a cli command such as `$auth config init` for what we're trying to accomplish from the config-generator? that command would skip all the shared config init (like `help`). The config generator ignores all the details. And we provide CLI users a way to do the same if they want to use the CLI from an automated scenario

internal/cli/config.go
```diff
@@ -0,0 +1,137 @@
...
+				return err
+			}
+
+			u, err := url.Parse("https://" + p.clientDomain)
+			if err != nil {
+				return err
+			}
+
+			// integration test client doesn't have openid or offline_access scopes granted
+			scopesForTest := requiredScopes[2:]
```
as-herzog, line 92: If we are providing this command for users, not sure we can do this anymore. Do we change the test client scopes or add a "scopes" env variable or something?

pkg/auth0-cli-config-generator/main.go
as-herzog: Took a shot at what I think you are suggesting, let me know what you think.

internal/cli/config.go
jfatta: besides that, I'll try to avoid doing `[2:]` it is heavily based on the order of the elements. Maybe we can handle this at expose this clearly from the auth package, like: scopes.All, scopes.Min, etc

internal/cli/config.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	cmd.AddCommand(initCmd(cli))
+	return cmd
+}
+
+func initCmd(cli *cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "initialize valid cli config from environment variables",
```
jfatta, line 73: I'd suggest
```suggestion
		Short: "configure the CLI from environment variables",
```

otherwise it is too internal

internal/cli/config.go
```diff
@@ -0,0 +1,137 @@
...
+Interactive logins should use "auth0 login" instead.`
+
+type params struct {
+	filePath     string
+	clientDomain string
+	clientID     string
+	clientSecret string
+}
+
+func (p params) validate() error {
```
jfatta, line 35: consider https://github.com/golang/go/wiki/CodeReviewComments#error-strings

tl;dr: start with lowercase and without end punctuation

pkg/auth0-cli-config-generator/main.go
jfatta: this is great. It's looking pretty good. I left some random comments. I can fully review it later, but thanks for doing this 🙇🏼 
integration tests are passing already using the command? 😮

pkg/auth0-cli-config-generator/main.go
```diff
@@ -1,224 +0,0 @@
-// auth0-cli-config-generator: A command that generates a valid config file that can be used with auth0-cli.
```
jfatta, line 1: it feels so good

jfatta (commented):
(just re-submitting because I received the review request, my comments from the prev review are still relevant)

jfatta (approved):
Looks good.

I'd suggest flagging the new command as [hidden](https://github.com/spf13/cobra/blob/6d00909120c77b54b0c9974a4e20ffc540901b98/command.go#L197) just to prevent unnecessary usage from end-users.

internal/cli/config.go
```diff
@@ -0,0 +1,134 @@
...
+				return fmt.Errorf("unexpected error adding tenant to config: %w", err)
+			}
+			return nil
+		},
+	}
+	viper.SetEnvPrefix("AUTH0_CLI")
+	viper.AutomaticEnv()
+
+	flags := cmd.Flags()
+	flags.String("filepath", path.Join(os.Getenv("HOME"), ".config", "auth0", "config.json"), "Filepath for the auth0 cli config")
```
Widcket, line 124: Could you please extract the path generation (`path.Join(os.Getenv("HOME"), ".config", "auth0", "config.json")`) into a method and use it in `setPath()`, `initContext()` and `main()`?

internal/cli/config.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	cmd.AddCommand(initCmd(cli))
+	return cmd
+}
+
+func initCmd(cli *cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "configure the CLI from environment variables",
```
Widcket, line 72:
```suggestion
		Short: "Configure the CLI from environment variables",
```

internal/cli/config.go
as-herzog: done

Widcket (approved):
Thank you!!

PR #293, opened 2021-05-14
feat: add roles

cyx:
## Description

This PR adds list, show, create, update, delete commands for the `role` resource.

A quick demo is shown below showcasing all the commands.

[asciicast](https://asciinema.org/a/Q2WDzn5LZQNgofaxyEHkMyJJ9)

### DIFF
--- a/internal/cli/roles.go
+++ b/internal/cli/roles.go
@@ -1,11 +1,37 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/auth0/auth0-cli/internal/ansi"
+	"github.com/auth0/auth0-cli/internal/prompt"
 	"github.com/spf13/cobra"
 	"gopkg.in/auth0.v5/management"
+)
+
+// errNoRoles signifies no roles exist in a tenant
+var errNoRoles = errors.New("there are currently no roles")
+
+var (
+	roleID = Argument{
+		Name: "Role ID",
+		Help: "Id of the role.",
+	}
+	roleName = Flag{
+		Name:       "Name",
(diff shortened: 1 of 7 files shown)


internal/display/roles.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (v *roleView) AsTableHeader() []string {
+	return []string{"Name", "Role ID", "Description"}
+}
+
+func (v *roleView) AsTableRow() []string {
+	return []string{
+		v.Name,
+		ansi.Faint(v.ID),
```
Widcket, line 21: Can we have the ID go first? Like in the other list commands.

internal/cli/roles.go
```diff
@@ -0,0 +1,303 @@
...
+	return cmd
+}
+
+func listRolesCmd(cli *cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Args:    cobra.NoArgs,
+		Short:   "List your roles",
+		Long: `List your existing roles. To create one try:
```
Widcket, line 61: Can you please run the file through `gofmt`?

Widcket (changes requested):
LGTM, just a couple of tiny details.

internal/display/roles.go
cyx: Yep, done

internal/cli/roles.go
cyx: Yeah, I think this is special gofmt logic.

Widcket (approved):

PR #294, opened 2021-05-14
fix: better development ergonomics

cyx:
When we're developing, we woudn't want to swallow panics since that can
cause a lot of surprises.

For non-development cases, since we're reporting this, then swallowing
the panic and displaying a basic error message would be enough for the
user to know that something went wrong happened.

Screen Shot 2021-05-14 at 4 13 20 PM


Widcket (approved):
Good idea, thanks!

PR #295, opened 2021-05-14
fix: remove stale actions code

cyx:
## Description

This is a continuation of #288 . For some reason, I thought that was done, but apparently not 😆


Widcket (approved):
Thanks for cleaning this up!

PR #296, opened 2021-05-14
fix: remove shim around lifetime in seconds

cyx:
Continuation of #276. Connects to https://github.com/go-auth0/auth0/pull/213


Widcket (approved):
Thanks!

PR #302, opened 2021-05-23
Add Heap integration [CLI-158]

Widcket:
### Description

This PR integrates Heap to track the following events (only on production builds):

- The first login
- Every command executed that requires login

Screen Shot 2021-05-23 at 18 29 54

Screen Shot 2021-05-23 at 18 28 42

### Testing

### Checklist


internal/analytics/analytics.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (t *Tracker) track(eventName string, id string) {
+	if !shouldTrack() {
+		return
+	}
+
+	event := newEvent(eventName, id)
+
+	t.wg.Add(1)
```
cyx, line 77: Just a commentary that we can't technically do `wg.Add` _after_ we do `wg.Wait` -- that would cause a panic. As long as we're synchronously done doing all the `wg.Add`s before we Wait, it's fine.

There's other async style ways we can do this as well (e.g. utilizing a single channel,  a single worker) to future proof against this potential issue.

cyx (approved):
Overall, just a minor potential issue with wg.Add / wg.Wait races, but 👍

PR #310, opened 2021-06-11
Replace client-id with filter in logs commands

Widcket:
### Description

This PR replaces the `--client-id` filter with a single `--filter` flag that takes a Lucene query string, like `users search`.

E.g.

```sh
auth0 logs list --filter "client_id:<client-id>"
auth0 logs list --filter "client_name:<client-name>"
auth0 logs list --filter "user_id:<user-id>"
auth0 logs tail --filter "user_name:<user-name>"
auth0 logs tail --filter "ip:<ip>"
auth0 logs tail --filter "type:f"
```

⚠️  **THIS IS A BREAKING CHANGE**

The alternative, which would have been to add a flag per valid filter field, would have meant having a bunch of flags that are mutually exclusive.

Screen Shot 2021-06-11 at 12 21 46

Screen Shot 2021-06-11 at 12 23 07

Screen Shot 2021-06-11 at 12 19 27

Screen Shot 2021-06-11 at 12 20 41

### Testing

### Checklist

### DIFF
--- a/docs/auth0_logs_list.md
+++ b/docs/auth0_logs_list.md
@@ -1,6 +1,6 @@
 ## auth0 logs list
 
-Show the tenant logs allowing to filter by Client Id.
+Show the tenant logs allowing to filter using Lucene query syntax.
 
 ```
 auth0 logs list [flags]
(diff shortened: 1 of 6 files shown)


cyx (approved):
Thanks!

PR #311, opened 2021-06-11
Add custom domains CRUD

Widcket:
### Description

- `auth0 branding domains list`
- `auth0 branding domains show`
- `auth0 branding domains create`
- `auth0 branding domains delete`
- `auth0 branding domains verify`

The `update` command cannot be added ATM because the Go SDK [does not support](https://github.com/go-auth0/auth0/blob/master/management/custom_domain.go) that operation.

Screen Shot 2021-06-11 at 19 35 15

Screen Shot 2021-06-11 at 19 35 26

### Testing

### Checklist


internal/cli/custom_domains.go
```diff
@@ -0,0 +1,319 @@
...
+		Long: `List your existing custom domains. To create one try:
+auth0 branding domains create`,
+		Example: `auth0 branding domains list
+auth0 branding domains ls`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var list []*management.CustomDomain
+
+			if err := ansi.Waiting(func() error {
+				var err error
+				list, err = cli.api.CustomDomain.List()
```
jfatta, line 81: I noticed we're not passing the context downstream, to handle cancelation properly, eventually support timeouts from the root CLI command.

this can be done with the option
https://github.com/go-auth0/auth0/blob/5e36f59f4a866da1c4858b650302c04235dc3508/management/management.go#L399-L404

the command holds its context at `cmd.Context()`

It seems we're not doing this in any of our API invocations, so feel free to ignore this for now, I can tackle the update on a separated PR.

internal/cli/custom_domains.go
```diff
@@ -0,0 +1,319 @@
...
+func showCustomDomainCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		ID string
+	}
+
+	cmd := &cobra.Command{
+		Use:   "show",
+		Args:  cobra.MaximumNArgs(1),
+		Short: "Show a custom domain",
+		Long:  "Show a custom domain.",
```
jfatta, line 104: I'm wondering if we really need `long` if its equal to `short`. do you know what's cobra doing differently with these values?

internal/cli/custom_domains.go
```diff
@@ -0,0 +1,319 @@
...
+auth0 branding domains ls`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var list []*management.CustomDomain
+
+			if err := ansi.Waiting(func() error {
+				var err error
+				list, err = cli.api.CustomDomain.List()
+				return err
+			}); err != nil {
+				return fmt.Errorf("An unexpected error occurred: %w", err)
```
jfatta, line 84: shouldn't use https://github.com/auth0/auth0-cli/blob/77c96d394def391a2ed90d94485634230140b955/internal/auth0/error.go#L5 (pkg/errors under the hood)?

internal/cli/custom_domains.go
```diff
@@ -0,0 +1,319 @@
...
+			if !cli.force && canPrompt(cmd) {
+				if confirmed := prompt.Confirm("Are you sure you want to proceed?"); !confirmed {
+					return nil
+				}
+			}
+
+			return ansi.Spinner("Deleting custom domain", func() error {
+				_, err := cli.api.CustomDomain.Read(url.PathEscape(inputs.ID))
+
+				if err != nil {
```
jfatta, line 217: is "not found" the only source of errors there?

internal/cli/log_streams.go
```diff
@@ -63,7 +63,7 @@ var (
 	httpContentFormat = Flag{
 		Name:         "HTTP Content Format",
 		LongForm:     "http-format",
-		Help:         "HTTP Content-Format header. Possible values: JSONLINES, JSONARRAY, JSONOBJECT.",
+		Help:         "HTTP Content-Format header. Possible values: jsonlines, jsonarray, jsonobject.",
```
jfatta, line 66: is this related with the domains CRUD?

internal/display/custom_domain.go
```diff
@@ -0,0 +1,97 @@
+package display
+
+import (
+	"strconv"
+
+	"github.com/auth0/auth0-cli/internal/ansi"
+	"gopkg.in/auth0.v5/management"
+)
+
+type customDomainView struct {
```
jfatta, line 10: what do you think about and  approach like this to simplify here and there?

```go
type customDomainView struct {
   raw *management.CustomDomain
}

func (v *customDomainView) AsTableRow() []string {
	return []string{
		ansi.Faint(v.raw.ID),
		v.raw.Domain,
		v.raw.Status,
	}
}
```

jfatta (commented):
great work, I left some comments inline

internal/cli/custom_domains.go
Widcket: Yes, that would need to be done in a separate PR.

internal/cli/custom_domains.go
Widcket: Ideally we'd like that to be abstracted away and just use a builder that wraps cobra's with our own conventions. But that needs to be part of a larger refactor effort. For now the focus is on just hitting the feature goals

internal/cli/custom_domains.go
Widcket: We're just using that method for panics for now.

internal/cli/log_streams.go
Widcket: Not, but found out about that, and seemed too small a change to create a separate PR for, and having to find a reviewer. Ideally under different circumstances, I'd have made that into a separate PR. If you'd prefer it that way I can extract it.

internal/cli/custom_domains.go
Widcket: We only do that GET to find out if the resource exists. Granted, it can fail because other reasons as well (e.g. network error), but unfortunately, we have no way to tell them apart. The delete call does not fail if the resource didn't exist, it ends successfully. That might be the worse outcome in this case.
What do you suggest?

internal/display/custom_domain.go
Widcket: I think keeping a separate view model that does not know about the underlying DTO is something that will make it easier and simpler to decouple the SDK in the future (if the project gets to that stage). Otherwise the data layer gets a (bigger) hard coupling with the view layer.

internal/cli/custom_domains.go
Widcket: Maybe rework the error message to mention that there are other potential reasons?

internal/display/custom_domain.go
Widcket: I mean, both layers are already coupled, but doing that will make it worse.

internal/cli/custom_domains.go
Widcket: (so that Sentry has a stack trace when it recovers from the panic)

internal/cli/custom_domains.go
jfatta: yes, maybe just remove the _The specified Id: %v doesn't exist_ and add the error msg?

jfatta (approved):

internal/cli/custom_domains.go
Widcket: Done in https://github.com/auth0/auth0-cli/pull/311/commits/8f932f292c4a17940a5545e89d4ef6539b544212

jfatta (approved):

PR #318, opened 2021-06-25
refactor: source auth vars from env

as-herzog:
### Description
Source auth variables from env if provided. This allows us to use the cli with cloud deployments. 

Screen Shot 2021-06-28 at 10 34 44 AM

Screen Shot 2021-06-28 at 10 34 17 AM


internal/auth/auth.go
```diff
@@ -9,21 +9,23 @@ import (
...
-	deviceCodeEndpoint     = "https://auth0.auth0.com/oauth/device/code"
-	oauthTokenEndpoint     = "https://auth0.auth0.com/oauth/token"
 	audiencePath           = "/api/v2/"
 	waitThresholdInSeconds = 3
-
 	// namespace used to set/get values from the keychain
 	SecretsNamespace = "auth0-cli"
 )
 
+var clientID = getEnv("CLIENT_ID", "2iZo3Uczt5LFHacKdM0zzgUO2eG2uDjT")
```
jfatta, line 24: we can simplify with github.com/joeshaw/envdecode
and a simple config struct in place. We could override any constant if needed, and also get rid of the logic to check the ENV.

```go
var cfg struct {
    ClientID            string `env:"AUTH0_CLIENT_ID,default=2iZo3Uczt5LFHacKdM0zzgUO2eG2uDjT"`
    BaseDomain     string `env:"AUTH0_BASE_DOMAIN,default=auth0.auth0.com"`
    DeviceCodeEndpoint     string `env:"AUTH0_DEVICE_CODE_ENDPOINT,default=https://%s/oauth/device/code"`
    OAuthTokenEndpoint     string `env:"AUTH0_OAUTH_TOKEN_ENDPOINT,default=https://%s/oauth/token"`
}

if err := envdecode.StrictDecode(&cfg); err != nil {
    return err
}
```

jfatta (commented):
great work, this is going to unlock custom deployments.

I left a suggestion to simplify the overall approach.

internal/auth/auth.go
as-herzog: Good call, I thought I remembered seeing this somewhere but couldn't find it 👍

internal/auth/auth.go
```diff
@@ -81,6 +68,29 @@ type State struct {
 	Interval        int    `json:"interval"`
 }
 
+func NewAuthenticaor(audience, id, deviceCodeUrl, tokenUrl string) Authenticator {
```
jfatta, line 71: there's no need for a New... method if internally is passing the received arguments directly, without required transformation, validation or initialization with default values. I'd suggest export (uppercase) those 4 properties (if needed) and use &Authenticator{...} where needed.

internal/auth/README.md
```diff
@@ -8,3 +8,14 @@ The CLI authentication follows this approach:
 1. During regular commands initialization, the access token is used to instantiate an Auth0 API client. 
 		- If the token is expired according to the value stored on the configuration file, a new one is requested using the refresh token. 
 		- In case of any error, the interactive login flow is triggered.
+
+
+### Customization
```
jfatta, line 13: if we keep the same on the higher level README I think it's enough

README.md
```diff
@@ -220,6 +220,17 @@ Failed Login	hello	7 minutes ago	N/A	my awesome app
 	userid: auth0|QXV0aDAgaXMgaGlyaW5nISBhdXRoMC5jb20vY2FyZWVycyAK
 ```
 
+## Customization
+
+The authenticator the CLI uses defaults to the default Auth0 cloud offering of `auth0.auth0.com`. This can be customized for personalized cloud offerings by setting the following env variables:
```
jfatta, line 225:
```suggestion
The authenticator of the CLI defaults to the default Auth0 cloud `auth0.auth0.com`. This can be customized for personalized cloud offerings by setting the following env variables:
```

internal/auth/token_test.go
```diff
@@ -46,8 +46,9 @@ func TestTokenRetriever_Refresh(t *testing.T) {
 		client := &http.Client{Transport: transport}
 
 		tr := &TokenRetriever{
-			Secrets: secretsMock,
-			Client:  client,
+			Authenticator: &Authenticator{"https://*.auth0.com/api/v2/", "2iZo3Uczt5LFHacKdM0zzgUO2eG2uDjT", "https://auth0.auth0.com/oauth/device/code", "https://auth0.auth0.com/oauth/token"},
```
jfatta, line 49: if we're not using the authenticator values during the tests, I'd suggest putting clearly dummy values here.

jfatta (approved):
I left minor suggestions, then I think is ready :shipit:

internal/auth/token_test.go
as-herzog: If I recall correctly the tests required correct URLs but we could probably dummy the id

as-herzog: ☹️ I didn't read the full test, fixed with dummy vals 👍

jfatta: thanks! I think it is clearer. If we need to update values in a future, this is clearly unrelated now 🙇🏼

Widcket (approved):
Awesome, thank you!

PR #320, opened 2021-06-30
Add `branding emails show` and `branding emails update` [CLI-190]

Widcket:
### Description

- `auth0 branding emails show`
- `auth0 branding emails update`

Screen Shot 2021-06-30 at 02 43 32

Screen Shot 2021-06-30 at 02 44 36

An issue with `actions update` was also fixed.

### Testing

### Checklist

### DIFF
--- a/internal/cli/email_templates.go
+++ b/internal/cli/email_templates.go
@@ -1,0 +1,292 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/auth0/auth0-cli/internal/ansi"
+	"github.com/spf13/cobra"
+	"gopkg.in/auth0.v5"
+	"gopkg.in/auth0.v5/management"
+)
+
+const (
+	emailTemplateVerifyLink     = "verify-link"
+	emailTemplateVerifyCode     = "verify-code"
+	emailTemplateChangePassword = "change-password"
+	emailTemplateWelcome        = "welcome"
+	emailTemplateBlockedAccount = "blocked-account"
+	emailTemplatePasswordBreach = "password-breach"
+	emailTemplateMFAEnrollment  = "mfa-enrollment"
+	emailTemplateMFACode        = "mfa-code"
+	emailTemplateUserInvitation = "user-invitation"
+)
+
+var (
+	emailTemplateTemplate = Argument{
+		Name: "Template",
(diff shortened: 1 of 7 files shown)


internal/cli/email_templates.go
```diff
@@ -0,0 +1,292 @@
...
+		{"Verification Email (using Link)", emailTemplateVerifyLink},
+		{"Verification Email (using Code)", emailTemplateVerifyCode},
+		{"Change Password", emailTemplateChangePassword},
+		{"Welcome Email", emailTemplateWelcome, },
+		{"Blocked Account Email", emailTemplateBlockedAccount},
+		{"Password Breach Alert", emailTemplatePasswordBreach},
+		{"Enroll in Multifactor Authentication", emailTemplateMFAEnrollment},
+		{"Verification Code for Email MFA", emailTemplateMFACode},
+		{"User Invitation", emailTemplateUserInvitation},
+	}
```
Widcket, line 95: These are the same strings used in Manage:

Screen Shot 2021-06-30 at 00 08 23

internal/cli/email_templates.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (c *cli) emailTemplateEditorHint() {
+	c.renderer.Infof("%s once you close the editor, the email template will be saved. To cancel, CTRL+C.", ansi.Faint("Hint:"))
+}
+
+func (c *cli) emailTemplatePickerOptions() (pickerOptions, error) {
+	return emailTemplateOptions, nil
+}
+
+func apiEmailTemplateFor(v string) string {
```
jfatta, line 269: this is hard work. I'm wondering how we're going to manage breaking changes regarding of the hardcoded values here

jfatta (approved):

internal/cli/email_templates.go
Widcket: We could deprecate values and still accept the old one (and use the new one internally) and print a warning. And eventually stop accepting the old value.
There are a couple of ways we could go about the last step:
- Remove support for deprecated values in new majors
- Establish a policy for deprecations, where a deprecated value is guaranteed to be accepted for x time, and then can be removed
- ... (suggestions?)

acoomans:
Hi, 
We have a script that would pipe in the body of the email template to `auth0 branding emails update`. It does not seem to work anymore. Did this change remove that functionality? I changed the script to use the `--body` parameter, but I'm worried of overflowing the command line. Are you planning on bringing piping back? thanks
Arnaud

PR #321, opened 2021-07-02
chore: update brew homepage

jfatta:
Update brew and scoop homepage to point to https://auth0.github.io/auth0-cli.
and other minor fixed to satisfy core brew.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -49,7 +49,7 @@ brews:
       name: auth0
       email: support@auth0.com
     homepage: https://auth0.github.io/auth0-cli
-    description: Auth0 command-line tool to supercharge your developer workflow.
+    description: Auth0 command-line tool to supercharge your developer workflow
     license: MIT
     skip_upload: auto
     install: |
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -48,8 +48,8 @@ brews:
     commit_author:
       name: auth0
       email: support@auth0.com
-    homepage: https://cli.auth0.com
-    description: Supercharge your developer workflow.
+    homepage: https://auth0.github.io/auth0-cli
+    description: Auth0 command-line tool to supercharge your developer workflow
```
jfatta, line 52: from brew tooling to verify a formula:
```
auth0/auth0-cli/auth0:
  * 5: col 1: A `test do` test block should be added
  * 6: col 44: Description shouldn't end with a full stop.
  * 42: col 16: Place the first line of a multi-line method definition's body on its own line.
Error: 3 problems in 1 formula detected

```

I'm still figuring out how to add the test from the goreleases definition. But we can merge this in the meantime.

Widcket (approved):

PR #324, opened 2021-07-02
docs: add favicon

jfatta:
Add auth0 favicon to docs homepage

image


docs/_includes/disqus_comments.html
```diff
@@ -0,0 +1,20 @@
+{%- if page.comments != false and jekyll.environment == "production" -%}
```
Widcket, line 1: Is this file necessary if we don't use Disqus?

docs/_includes/footer.html
```diff
@@ -0,0 +1,35 @@
+<footer class="site-footer h-card">
```
Widcket, line 1: Why are we adding a custom footer?

docs/_includes/google-analytics.html
```diff
@@ -0,0 +1,12 @@
+<script>
```
Widcket, line 1: Should we add this?

docs/_includes/head.html
```diff
@@ -0,0 +1,12 @@
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  {%- seo -%}
+  <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
+  <link rel="icon" type="image/png" href="favicon.png">
+  {%- feed_meta -%}
+  {%- if jekyll.environment == 'production' and site.google_analytics -%}
+    {%- include google-analytics.html -%}
```
Widcket, line 10: Same as above, do we need GA?

docs/_includes/social.html
```diff
@@ -0,0 +1,14 @@
+<ul class="social-media-list">
```
Widcket, line 1: What are these for? Do we want this?

Widcket: Same for the above individual social icons.

jfatta: @Widcket, yes. To override **one** include, we need to respect the entire folder structure here https://github.com/jekyll/minima/tree/master/_includes. It is a common practice with github themes until you have your own. I tried adding only the header to our _includes folder and the `make start-doc` process fails.

Widcket (approved):

PR #329, opened 2021-07-09
feat: add role permissions

cyx:
### Description

This feature adds a set of commands:

```
auth0 roles permissions list
auth0 roles permissions associate
auth0 roles permissions remove
```

### DIFF
--- a/internal/cli/roles_permissions.go
+++ b/internal/cli/roles_permissions.go
@@ -30,15 +30,14 @@ var (
 
 func rolePermissionsCmd(cli *cli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "permissions",
-		Short:   "Manage permissions within the role resource",
-		Long:    "Manage permissions within the role resource.",
-		Aliases: []string{"permission"},
+		Use:   "permissions",
+		Short: "Manage permissions within the role resource",
+		Long:  "Manage permissions within the role resource.",
 	}
 
 	cmd.SetUsageTemplate(resourceUsageTemplate())
 	cmd.AddCommand(listRolePermissionsCmd(cli))
-	cmd.AddCommand(associateRolePermissionsCmd(cli))
+	cmd.AddCommand(addRolePermissionsCmd(cli))
 	cmd.AddCommand(removeRolePermissionsCmd(cli))
 
 	return cmd
@@ -55,7 +54,7 @@ func listRolePermissionsCmd(cli *cli) *cobra.Command {
 		Args:    cobra.MaximumNArgs(1),
 		Short:   "List permissions defined within a role",
 		Long: `List existing permissions defined in a role. To add a permission try:
-auth0 roles permissions associate <role-id>`,
+auth0 roles permissions add <role-id>`,
 		Example: `auth0 roles permissions list <role-id>
 auth0 roles permissions ls`,
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -86,7 +85,7 @@ auth0 roles permissions ls`,
 	return cmd
 }
 
-func associateRolePermissionsCmd(cli *cli) *cobra.Command {
+func addRolePermissionsCmd(cli *cli) *cobra.Command {
 	var inputs struct {
 		ID            string
 		APIIdentifier string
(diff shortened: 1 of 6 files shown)


internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,254 @@
...
+		IsRequired: true,
+	}
+)
+
+func rolePermissionsCmd(cli *cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "permissions",
+		Short:   "Manage permissions within the role resource",
+		Long:    "Manage permissions within the role resource.",
+		Aliases: []string{"permission"},
```
Widcket, line 36: Why that alias in particular? We don't have any other aliases (nor commands)  in singular.

internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/auth0/auth0-cli/internal/ansi"
+	"github.com/spf13/cobra"
+	"gopkg.in/auth0.v5/management"
+)
+
+var (
+	roleAPIIdentifier = Flag{
+		Name:       "API",
+		LongForm:   "api-id",
```
Widcket, line 16: This should just probably be an argument, like the other subresources in the CLI:
- `auth0 users blocks list <user-id>`
- `auth0 apis scopes list <id|audience>`

internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,254 @@
...
+
+func associateRolePermissionsCmd(cli *cli) *cobra.Command {
+	var inputs struct {
+		ID            string
+		APIIdentifier string
+		Permissions   []string
+	}
+
+	cmd := &cobra.Command{
+		Use:     "associate",
```
Widcket, line 97: What about finding shorter synonym? E.g manage uses `add`. This would remove the need for an alias.

Screen Shot 2021-07-09 at 23 30 47

Widcket: Also, it would go well with `remove` below.

internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,254 @@
...
+			ps := makePermissions(rs.GetIdentifier(), inputs.Permissions)
+			if err := cli.api.Role.AssociatePermissions(inputs.ID, ps); err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	roleAPIIdentifier.RegisterString(cmd, &inputs.APIIdentifier, "")
```
Widcket, line 140: As stated above, this should probably be just an arg.

internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,254 @@
...
+			ps := makePermissions(rs.GetIdentifier(), inputs.Permissions)
+			if err := cli.api.Role.RemovePermissions(inputs.ID, ps); err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	roleAPIIdentifier.RegisterString(cmd, &inputs.APIIdentifier, "")
```
Widcket, line 196: Same as above.

internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,254 @@
...
+	var options []string
+	for _, s := range rs.Scopes {
+		options = append(options, s.GetValue())
+	}
+
+	prompt := &survey.MultiSelect{
+		Message: "Permissions",
+		Options: options,
+	}
+	survey.AskOne(prompt, permissions)
```
Widcket, line 240: We could use our own wrapper, `prompts.AskOne` that has the correct settings and such.

Widcket: And add a new input there, `MultiSelectInput`.

internal/cli/roles_permissions.go
cyx: The difference here is we need two separate things:

1. The role ID
2. On top of that, the api-id.

Do you mean to say two arguments?

internal/cli/roles_permissions.go
cyx: `prompts.AskOne` doesn't work with a `MultiSelect`.

internal/cli/roles_permissions.go
Widcket: Oh, you're right 🤦🏼‍♀️

internal/cli/roles_permissions.go
```diff
@@ -0,0 +1,268 @@
...
+	cmd := &cobra.Command{
+		Use:   "add",
+		Args:  cobra.MaximumNArgs(1),
+		Short: "Add a permission to a role",
+		Long: `Add an existing permission defined in one of your APIs.
+To add a permission try:
+
+    auth0 roles permissions add <role-id> -p <permission-name>`,
+		Example: `auth0 roles permissions add <role-id> -p <permission-name>
+auth0 roles permissions assoc`,
```
Widcket, line 105: The example needs to be updated

internal/display/role_permissions.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (r *Renderer) RolePermissionList(perms []*management.Permission) {
+	resource := "role permissions"
+
+	r.Heading(resource)
+
+	if len(perms) == 0 {
+		r.EmptyState(resource)
+		r.Infof("Use 'auth0 roles permissions associate' to add one")
```
Widcket, line 51:
```suggestion
		r.Infof("Use 'auth0 roles permissions add' to add one")
```

internal/cli/roles_permissions.go
Widcket: It still says `assoc`, as there's no alias now, the second example should be removed.

cyx: Sorry missed that. Done!

Widcket (approved):
Awesome, thanks!

PR #330, opened 2021-07-10
Refactor editor prompt to unblock tests

Widcket:
### Description

Currently, it is not possible to add integration tests for any command that involves opening an external code editor, such as `rules create` or `rules update`. The editor prompt does not honor the `--no-input` flag, and we'd also need to add support for an alternative input method (e.g. piping).

This PR tackles the first problem, by having the editor prompt use the same code infrastructure as the rest of the prompts. That makes the editor behave similarly to the other prompts, e.g honoring `--no-input` and handling errors the same way.

### Testing

### Checklist


cyx (approved):
Looks great! Is there a UX change or is it completely transparent?

Widcket:
It is transparent.

PR #331, opened 2021-07-10
Fix askManyFlag

Widcket:
### Description

This PR fixes the logic in `askManyFlag` that required an extra check when using a slice flag.

Screen Shot 2021-07-09 at 23 14 04
Screen Shot 2021-07-09 at 23 13 25

### Testing

### Checklist


jfatta (approved):
looks good, thanks

PR #358, opened 2021-09-21
Removed role alias

Widcket:
### Description

This PR removes the `role` alias from the `roles` command, as the `roles` command is already short enough and an alias with  1 character less is not justified, and does not follow from the other aliases:

Screen Shot 2021-09-21 at 17 13 44

**THIS IS A BREAKING CHANGE**.

### Testing

### Checklist


jfatta (approved):
makes sense 👍🏼

PR #384, opened 2022-01-18
Add homebrew docs for Linux

Widcket:
### Description

This PR adds instructions to the README to install the CLI using Homebrew on Linux, given that the formula is available for Linux installs as well: https://github.com/auth0/homebrew-auth0-cli/blob/main/auth0.rb#L38

### Testing

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,12 @@ scoop install auth0
 
 ### Linux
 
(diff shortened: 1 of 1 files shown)


jfatta (approved):
🐧 nice

PR #387, opened 2022-01-21
Add `actions deploy` command [DXCDT-13]

### LINKED ISSUE #352: Deploy actions once created
### Describe the problem you'd like to have solved

CLI allows to create an action and even specify a trigger, but it's not deployed (and there seems to be no command or flag to deploy via CLI at the moment). This means that customizing the actions still requires two manual actions on the Auth0 portal afterwards: deploying the action and then adding it to the flow.

### Describe the ideal solution

In order to fully support CLI-based provisioning of the tenants, it would be great to support deploying the action from CLI and/or adding it to the flow (not sure if the latter is feasible, given that other actions might already be there and that the order probably matters).

### Additional context

My main use case is provisioning the unified set of roles and permissions on multiple tenants corresponding to the deployment stages (dev, test, demo, production etc.). Our ideal goal would be to configure a tenant in a fully automated way: define API, permissions, standard roles, custom login actions - in essence, make tenant configuration reproducible, minimize manual work, and reduce human errors.

Are there any plans to support that or good reasons not to?

### Technical considerations

From looking at the Management API, I can see that [Deploy and action](https://auth0.com/docs/api/management/v2/#!/Actions/post_deploy_action) and [Update trigger bindings](https://auth0.com/docs/api/management/v2/#!/Actions/patch_bindings) methods would essentially allow to perform these steps, so it is at least theoretically possible.

Widcket:
### Description

This PR adds support for deploying actions via the `auth0 actions deploy` command.

Screen Shot 2022-01-20 at 22 02 38

### References

Closes #352 

### Testing

The changes were tested manually.

### Checklist

### DIFF
--- a/internal/cli/actions.go
+++ b/internal/cli/actions.go
@@ -78,6 +78,7 @@ func actionsCmd(cli *cli) *cobra.Command {
 	cmd.AddCommand(showActionCmd(cli))
 	cmd.AddCommand(updateActionCmd(cli))
 	cmd.AddCommand(deleteActionCmd(cli))
+	cmd.AddCommand(deployActionCmd(cli))
 	cmd.AddCommand(openActionCmd(cli))
 
 	return cmd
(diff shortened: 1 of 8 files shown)


internal/cli/actions.go
```diff
@@ -138,7 +139,7 @@ auth0 actions show <id>`,
 
 			if err := ansi.Waiting(func() error {
 				var err error
-				action, err = cli.api.Action.Read(url.PathEscape(inputs.ID))
```
Widcket, line 141: `url.PathEscape` is not necessary for action IDs, except when building the URL for `auth0 actions open`.

sergiughf (approved):

PR #388, opened 2022-01-21
Hide global `--tenant` flag on login command [DXCDT-14]

### LINKED ISSUE #372: auth-cli login doesnt respect tenant
When doing `auth0 login --tenant mytenant-dev.eu.auth0.com` there's no guarantee that that is the tenant I will be authenticated against - it seems to be completely random if I have access to multiple tenants in my browser.

I would expect the login command to fail if I end up authenticated to a different tenant than specified - the current functionality is almost a security risk.

I've posted a couple of other issues regarding auth0-cli usage in a multi-tenant situation and I'm afraid the cli is not usable in it's current form for us because of these issues.

Widcket:
### Description

This PR hides the `--tenant` global flag from the help of the `auth0 login` command.

The purpose of the --tenant flag is to allow CLI users to pick a tenant they’re already logged into to perform a specific operation – not to pick the tenant to log in to. This can easily be a source of confusion.

**Marked as draft until #387 gets merged because we need to generate the updated docs, and that PR also includes docs.**

#### Before

Screen Shot 2022-01-21 at 13 38 33

#### After

Screen Shot 2022-01-21 at 13 35 51

### References

Closes #372 

### Testing

The change was tested manually.

### Checklist


sergiughf (approved):
Looks good! Thanks @Widcket !

PR #390, opened 2022-01-21
Have the login command honor the `--no-input` flag [DXCDT-15]

### LINKED ISSUE #371: print the sign-in url to make it easier to sign in in a different browser
when running `auth0 login` it's too hard to use a non-default browser to sign in. It would help a lot if the cli printed the sign-in url so that one could paste it into the correct browser. I'm a little surprised that not even the `--debug` option seems to print this info.

Widcket:
### Description

The `--no-input` flag is meant to disable interactivity. This PR has the `auth0 login` command honor this flag so as to not automatically open the login URL in the default browser, but print it instead. This allows the use of any browser for login.

Screen Shot 2022-01-21 at 18 45 44

The regular, interactive login continues to work normally:

Screen Shot 2022-01-21 at 18 38 01

### References

Fixes #371

### Testing

The change was tested manually.

### Checklist

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -8,8 +8,8 @@ import (
 	"github.com/auth0/auth0-cli/internal/ansi"
 	"github.com/auth0/auth0-cli/internal/auth"
 	"github.com/auth0/auth0-cli/internal/prompt"
-	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


sergiughf (approved):
This is great! Thanks @Widcket!

Issue #392, opened 2022-02-02
The process to use this tool on a private cloud is unclear


andrewclaus:
### Description

The readme should clarify how to use this tool for private cloud instances.

Currently, the readme has this to say:

```
The authenticator of the CLI defaults to the default Auth0 cloud auth0.auth0.com. This can be customized for personalized cloud offerings by setting the following env variables:

	AUTH0_AUDIENCE - The audience of the Auth0 Management API (System API) to use.
	AUTH0_CLIENT_ID - Client ID  of an application configured with the Device Code grant type.
	AUTH0_DEVICE_CODE_ENDPOINT - Device Authorization URL
	AUTH0_OAUTH_TOKEN_ENDPOINT - OAuth Token URL
```

When I set these 4 env variables and run `auth0 login`, I am redirected to the finder app (I am running on a Mac) instead of the web browser. It is possible this is because I have the wrong values set and it is unable to start the device code process. Alternatively, there could be a bug with auth0-cli that will always redirect to finder if AUTH0_CLIENT_ID is set. See the conversation here: https://github.com/auth0/auth0-cli/issues/354

If I have the wrong values set, then it might be a good idea to clarify in the readme where these values can be found. 

### Alternative

Through trial and error, I did manage to get the CLI working by running these commands:

```
export AUTH0_CLI_CLIENT_DOMAIN=TENANT.PRIVATE_INSTANCE.auth0.com
export AUTH0_CLI_CLIENT_ID=CLIENT_ID_OF_M2M_APPLICATION_IN_TENANT
export AUTH0_CLI_CLIENT_SECRET=SECRET_OF_M2M_APPLICATION_IN_TENANT
auth0 config init
```

This created ~.config/auth0/config.json and I was then able to run additional auth0 commands!

If this is a valid method of setting this up, then I would suggest adding it to the readme.

Widcket:
Hi @andrewclaus, thanks for raising this.

The process documented in the README is the one intended for getting the interactive login to work (`auth0 login`) with a private cloud instance. As you mention, some users are getting an error when the `AUTH0_CLIENT_ID` is set, and we'll be looking into this.

The other process is for getting the CLI to work in non-interactive environments like a CI job, but is a valid workaround for the previous issue. We could document the non-interactive process in the README as well.

I'll leave this issue open to track that.

willvedd:
[...]

PR #394, opened 2022-02-05
fix: auth0 tenants add: blank slate

cyx:
### Description

For the blank slate case, we don't want it to error out if there are no
tenants since by definition we have no tenants configured.

### References

https://github.com/auth0/auth0-cli/issues/392

### Testing

1. Remove all your config e.g. `rm -rf ~/.config/auth0`
2. Try doing `auth0 tenants add`

You should be prompted, instead of seeing an error.

### Checklist

### DIFF
--- a/internal/cli/tenants.go
+++ b/internal/cli/tenants.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/auth0/auth0-cli/internal/prompt"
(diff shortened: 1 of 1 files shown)


internal/cli/tenants.go
```diff
@@ -166,7 +167,13 @@ func addTenantCmd(cli *cli) *cobra.Command {
 			if len(args) == 0 {
 				err := tenantDomain.Pick(cmd, &inputs.Domain, cli.tenantPickerOptions)
 				if err != nil {
-					return err
+					if !errors.Is(err, errUnauthenticated) {
+						return err
+					}
```
cyx, line 172: In the event we get a different type of error we won't silence it accidentally. For this case, we only care about `errUnauthenticated`.

internal/cli/tenants.go
```diff
@@ -166,7 +167,13 @@ func addTenantCmd(cli *cli) *cobra.Command {
...
+
+					if err := tenantDomain.Ask(cmd, &inputs.Domain); err != nil {
+						return err
+					}
```
cyx, line 176: We revert to a basic prompt since there's no "list" of tenants yet to formulate / present.

Widcket (commented):
Thank you!

Widcket (approved):

PR #401, opened 2022-03-09
Add JSON config reader

evansims:
@Widcket Took a stab at adding the config reader, although my inexperience with Golang is showing, and I'm not quite sure how to invoke it from within `importCmd` in `internal/cli/import.go` at present 😆


internal/cli/import/config.go
```diff
@@ -1,8 +1,29 @@
...
-// Put here the logic for parsing the JSON config file
+import (
+	"encoding/json"
+	"io/ioutil"
+	"log"
+)
 
 type Config struct {
 	Auth0Domain                 string   `json:"AUTH0_DOMAIN"`
 	Auth0KeywordReplaceMappings struct{} `json:"AUTH0_KEYWORD_REPLACE_MAPPINGS"`
```
Widcket, line 11: This should probably be a `map[string]interface{}`, because it can contain any valid JSON value (e.g. numbers, strings). See https://auth0.com/docs/deploy-monitor/deploy-cli-tool/environment-variables-and-keyword-mappings, e.g.
```json
"AUTH0_KEYWORD_REPLACE_MAPPINGS": {
  "ENVIRONMENT_URL": "http://dev.travel0.com",
  "JWT_TIMEOUT": 120,
  ...
}
```

Widcket (changes requested):
The logic for unmarshalling is Ok, but `Auth0KeywordReplaceMappings` should be a map of empty interfaces.

Widcket:
In golang, the empty interface `interface{}` basically stands in for any, and needs to be cast.

Widcket:
Also, please add `AUTH0_ALLOW_DELETE` to the config, as a string.

Widcket:
Checked out the PR to see how to use the method:
Please rename the `import` folder to `importcmd`, otherwise it conflicts with the `import` keyword.

Screen Shot 2022-03-09 at 18 50 59

Then in every file inside it, replace `package cli` with `package importcmd`.
Screen Shot 2022-03-09 at 18 51 04

Finally, PascalCase the name of the method to export it:
Screen Shot 2022-03-09 at 18 52 37

internal/cli/import/config.go
```diff
@@ -1,8 +1,29 @@
...
 }
+
+func getConfig(path string) *Config {
```
Widcket, line 14: This should return the error as well, instead of printing it: 
Screen Shot 2022-03-09 at 18 54 48
The error always goes last.

evansims:
@Widcket Thanks for the guidance! Learning a lot here 😆 Think I got everything you suggested knocked out here. Let me know your thoughts. (Build errors just seem to be around unused vars, as things aren't wired up yet.)

internal/cli/importcmd/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+)
+
+type Config struct {
+	Auth0Domain                 string                 `json:"AUTH0_DOMAIN"`
+	Auth0KeywordReplaceMappings map[string]interface{} `json:"AUTH0_KEYWORD_REPLACE_MAPPINGS"`
+	Auth0AllowDelete            string                 `json:"AUTH0_ALLOW_DELETE"`
```
Widcket, line 11:
```suggestion
	Auth0AllowDelete            bool                   `json:"AUTH0_ALLOW_DELETE"`
```
My bad 🙈 , this should be a bool.

Widcket (approved):

PR #404, opened 2022-03-11
Add handlers

Widcket:
### Description

This PR adds implementations for the apps, APIs, and roles handlers. Conflict resolution was not implemented.

https://user-images.githubusercontent.com/5055789/157835567-64759281-21bd-4556-85c3-ec93c5ad7918.mov

### Testing

### Checklist


internal/cli/import_yaml.go
```diff
@@ -0,0 +1,221 @@
...
+			errMsg := fmt.Sprintf("found duplicate name in resourceServer: %s", rs.Name)
+			result = multierror.Append(result, errors.New(errMsg))
+		}
+		resourceServerNames[rs.Name] = true
+	}
+
+	roleNames := make(map[string]bool)
+	for _, r := range t.Roles {
+		if roleNames[r.Name] {
+			errMsg := fmt.Sprintf("found duplicate name in resourceServer: %s", r.Name)
```
christonog, line 214: Probably missed by me earlier, but should change to `found duplicate name in role`
```suggestion
			errMsg := fmt.Sprintf("found duplicate name in role: %s", r.Name)
```

Widcket: Thank you!

internal/cli/import_yaml.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+
+		for j, item := range client.AllowedOrigins {
+			if strings.ContainsAny(item, key) {
+				str := strings.ReplaceAll(item, fmt.Sprintf("##%s##", key), replacement.(string))
+				t.Clients[i].AllowedOrigins[j] = str
+			}
+		}
+
+		for j, item := range client.GrantTypes {
```
christonog, line 130: I see we added more fields for clients and apis. Do we know what fields are configurable? Or can we allow replacements for all fields?

internal/cli/import_roles.go
```diff
@@ -0,0 +1,221 @@
...
+// 4. Perform patch diff
+// 5. Execute creates, updates and deletes
+
+// 1. Fetch all existing items
+
+func (cli *cli) getAllRoles(context context.Context) ([]*Role, error) {
+	list, err := getWithPagination(
+		context,
+		0, // Get *all* roles
+		func(opts ...management.RequestOption) (result []interface{}, hasNext bool, apiErr error) {
```
christonog, line 28: looks like `opts` is available with the context that is passed in?

internal/cli/import_yaml.go
Widcket: Thanks for pointing this out. I compared with the Management API docs, and ended adding `omitempty` for `app_type` (apps) ([commit](https://github.com/auth0/auth0-cli/pull/404/commits/01280ba29f28d6a86c0af685ee724dbbc0df9d05)), `name` (APIs) ([commit](https://github.com/auth0/auth0-cli/pull/404/commits/5387d55e3faf9d1b5faaa1a183b812a265922b09)), and `description` (roles) ([commit](https://github.com/auth0/auth0-cli/pull/404/commits/f50b7fd53224046e40fc202f5d82a8a771dea1f8)).

**Apps:** only `name` is required.
Screen Shot 2022-03-11 at 12 31 06

**APIs:** only `identifier` is required.
Screen Shot 2022-03-11 at 12 31 50

**Roles:** only `name` is required.
Screen Shot 2022-03-11 at 12 32 15

internal/cli/import_roles.go
christonog: I guess I'm curious where opts is passed in to `cli.getAllAPIs(ctx)` since only a ctx is passed in?

christonog (approved):
@Widcket Do you think you could share your `tenant.yaml` and non-secret `config.json` so I can give it a go on my end as well? Curious to see how it looks in the management dashboard.

Widcket:
> @Widcket Do you think you could share your `tenant.yaml` and non-secret `config.json` so I can give it a go on my end as well? Curious to see how it looks in the management dashboard.

Sure, I'll send via PM.

internal/cli/import_roles.go
Widcket: Oh, the func takes opts but none are passed of course. This is copypasta from https://github.com/auth0/auth0-cli/blob/main/internal/cli/organizations.go#L688.

PR #450, opened 2022-06-11
Add support for listing more than 50 roles

sergeybykov:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/.github/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

Add support for listing more than 50 roles in a tenant by using pagination.

### References

Fixes #449 

### Testing

Manual testing against a tenant with more than 50 roles.

- [ ] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [X] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [X] All active GitHub checks for tests, formatting, and security are passing
- [X] The correct base branch is being used, if not `master`

PR #457, opened 2022-06-28
feat: paginate apis list

Iristyle:
- Use prior art to add pagination around the apis list call.
   At present, there is no way to return more than the initial result of
   apis, which makes this command useless for any more than the default
   list length.

   Query functionality would be better, but this improves the situation
   by allowing for processing a larger group size

   Default list size is 50 items, like other resources

### Testing

> Describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors.
>
> Please include any manual steps for testing end-to-end or functionality not covered by unit/integration tests.
>
> Also include details of the environment this PR was developed in (language/platform/browser version).

### Checklist


sergiughf (approved):
Really appreciate the contribution @Iristyle! I pushed some small changes to your PR just so we have it consistent with how pagination was implemented in https://github.com/auth0/auth0-cli/pull/450 for roles, in order to save you some time.

Iristyle:
Awesome - thanks @sergiughf !

This will considerably help out with some cleanup automation

PR #532, opened 2022-11-29
DXCDT-283: Remove `config` command

willvedd:
### 🔧 Changes

Removes the `config` command. This command was undocumented and only existed to instantiate a tenant for integration tests but this functionality already exists with the `auth0 tenants add` command. Further, it was undocumented and not meant to be utilized by users. 

This PR accomplishes a few things:
- Have our integration tests follow the actual, idiomatic authentication mechanisms 
- Set us up for a cleaner transition when `auth0 login` and `auth0 tenants add` commands are consolidated
- Simplify the codebase

Originally, this command was added with #291 , at a time when `auth0 tenants add` did not get added until #304. 

### 📚 References

- Original `config` command PR: #291
- Tenants add command PR: #304

### 🔬 Testing

Confirms working if passes integration tests in CI.

### 📝 Checklist

- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [x] I have added documentation for all new/changed functionality (or N/A)

PR #537, opened 2022-12-01
DXCDT-293: Access token management for client credentials

willvedd:
### 🔧 Changes

The `api` command requires an access token to make authenticated management API requests. However, access tokens are not created for tenants authenticated via client credentials, only those authenticated via device code flow. This PR introduces the facilitation of access token management for tenants that are authenticated with client credentials.

Immediately, this work enables the above-mentioned `api` compatibility, however the intent is broader. This PR retrieves and stores the access token with client credentials. But more importantly, it standardizes both authentication processes and reduces the number of forks in the code to accommodate the two authentication mechanisms. but also lays the groundwork for code simplifications to occur later in time.

### 🔬 Testing

No tests created directly in this PR, IMO there is insufficient framework to facilitate them. However, the mechanism added will be tested before the running of all integration tests once #532 gets merged in.

### 📝 Checklist

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/joeshaw/envdecode"
+	"golang.org/x/oauth2/clientcredentials"
 )
 
 const (
@@ -65,7 +66,7 @@ type Result struct {
 	Domain       string
 	RefreshToken string
 	AccessToken  string
-	ExpiresIn    int64
+	ExpiresAt    time.Time
 }
 
 type State struct {
@@ -167,10 +168,14 @@ func (a *Authenticator) Wait(ctx context.Context, state State) (Result, error) {
 				return Result{}, fmt.Errorf("cannot parse tenant from the given access token: %w", err)
 			}
 
+			expiresAt := time.Now().Add(
+				time.Duration(res.ExpiresIn) * time.Second,
+			)
+
 			return Result{
 				RefreshToken: res.RefreshToken,
 				AccessToken:  res.AccessToken,
-				ExpiresIn:    res.ExpiresIn,
+				ExpiresAt:    expiresAt,
 				Tenant:       ten,
 				Domain:       domain,
 			}, nil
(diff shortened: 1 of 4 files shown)


internal/auth/auth.go
```diff
@@ -65,7 +66,7 @@ type Result struct {
 	Domain       string
 	RefreshToken string
 	AccessToken  string
-	ExpiresIn    int64
+	ExpiresAt    time.Time
```
willvedd, line 69: Making this change enabled me to reuse the `Result` type for client credential access tokens. The associated arithmetic gets slightly shuffled as a result.

internal/auth/auth.go
```diff
@@ -249,3 +254,39 @@ func parseTenant(accessToken string) (tenant, domain string, err error) {
 	}
 	return "", "", fmt.Errorf("audience not found for %s", audiencePath)
 }
+
+// ClientCredentials encapsulates all data to facilitate access token creation with client credentials (client ID and client secret)
+type ClientCredentials struct {
```
willvedd, line 259: Would've used the `clientcredentials.Config` type but I wanted to encapsulate the token URL formulation logic within this function.

internal/cli/cli.go
```diff
@@ -138,17 +138,10 @@ func (c *cli) setup(ctx context.Context) error {
...
-			management.WithClientCredentials(t.ClientID, t.ClientSecret),
-			management.WithUserAgent(ua),
-		)
-	} else {
-		m, err = management.New(t.Domain,
-			management.WithStaticToken(t.AccessToken),
-			management.WithUserAgent(ua),
-		)
-	}
+	m, err = management.New(t.Domain,
```
willvedd, line 141: This is one of the major simplifications that occurs when we rely simply on the access token for both authentication methods.

internal/cli/cli.go
```diff
@@ -168,27 +161,47 @@ func (c *cli) prepareTenant(ctx context.Context) (Tenant, error) {
...
 		if err != nil {
 			return Tenant{}, err
 		}
-	} else if isExpired(t.ExpiresAt, accessTokenExpThreshold) {
-		// check if the stored access token is expired:
-		// use the refresh token to get a new access token:
+		return t, nil
+	}
+
+	if isExpired(t.ExpiresAt, accessTokenExpThreshold) {
```
willvedd, line 172: Much of this diff is due to the elimination of the `else` statement in favor of an early return.

internal/cli/cli.go
```diff
@@ -168,27 +161,47 @@ func (c *cli) prepareTenant(ctx context.Context) (Tenant, error) {
...
+				AccessToken: token.AccessToken,
+				ExpiresAt:   token.ExpiresAt,
+				Scopes:      auth.RequiredScopes(),
+			}
+
+			if err := c.addTenant(t); err != nil {
+				return t, fmt.Errorf("unexpected error adding tenant to config: %w", err)
+			}
+			return t, nil
+		}
```
willvedd, line 196: This is how we "refresh" expired access tokens with client credentials.

internal/cli/cli.go
```diff
@@ -197,18 +210,19 @@ func (c *cli) prepareTenant(ctx context.Context) (Tenant, error) {
...
+		t.AccessToken = res.AccessToken
+		t.ExpiresAt = time.Now().Add(
+			time.Duration(res.ExpiresIn) * time.Second,
+		)
+
+		err = c.addTenant(t)
+		if err != nil {
+			return Tenant{}, err
 		}
+
```
willvedd, line 225: Similar to the above, this is a stylistic change to eliminate the `else` in favor of an early return.

sergiught (approved):
We reviewed this together with @willvedd and also performed manual tests too on both device code flow and client credentials. We deferred adding some automated unit tests due to the amount of changes needed at the moment to properly mock a couple of components.

PR #546, opened 2022-12-06
DXCDT-267: Consolidate `auth0 add tenants` into `auth0 login` (1/x)

willvedd:
### 🔧 Changes
There are two ways to authenticate, either as a user or as a machine (client credentials), which is currently accomplished by `auth0 login` and `auth0 tenants add` respectively. Having two widely different commands for authenticating has been confusing and not self-documenting. In an effort to simplify the process of authenticating with an Auth0 tenant, the two commands are being consolidated into the `auth0 login` command.

The major difference is the addition of a `--as-machine` flag which will indicate to the CLI to initiate the client credentials authentication process, otherwise the command remains unchanged.

**Many more improvements can be made to this command**, however, the size of this PR is already fairly large, so those will be coming in a series of follow-up PRs.

### 🔬 Testing

One additional integration test has been added, which will likely do for now. However, this begs the need for a more comprehensive integration testing solution that can better assert interactive behavior.

### 📝 Checklist


internal/cli/login.go
```diff
@@ -141,3 +185,41 @@ func RunLogin(ctx context.Context, cli *cli, expired bool) (Tenant, error) {
...
+func RunLoginAsMachine(ctx context.Context, inputs LoginInputs, cli *cli, cmd *cobra.Command) error {
+	if err := loginTenantDomain.Ask(cmd, &inputs.Domain, nil); err != nil {
+		return err
+	}
+
+	if err := tenantClientID.Ask(cmd, &inputs.ClientID, nil); err != nil {
+		return err
+	}
+
+	if err := tenantClientSecret.AskPassword(cmd, &inputs.ClientSecret, nil); err != nil {
```
Widcket, line 199: If this is intended for machine usage (automation scenarios where stdout won't be connected to a TTY), maybe we don't need to ask for the values?

Makefile
```diff
@@ -129,7 +129,7 @@ test-unit: ## Run unit tests
 test-integration: $(GO_BIN)/commander ## Run integration tests. To run a specific test pass the FILTER var. Usage: `make test-integration FILTER="attack protection"`
 	${call print, "Running integration tests"}
 	@$(MAKE) install # ensure fresh install prior to running test
-	auth0 tenants add ${AUTH0_CLI_CLIENT_DOMAIN} --client-id ${AUTH0_CLI_CLIENT_ID} --client-secret ${AUTH0_CLI_CLIENT_SECRET}  && commander test ./test/integration/test-cases.yaml --filter "$(FILTER)"; \
+	auth0 login --as-machine --domain ${AUTH0_CLI_CLIENT_DOMAIN} --client-id ${AUTH0_CLI_CLIENT_ID} --client-secret ${AUTH0_CLI_CLIENT_SECRET}  && commander test ./test/integration/test-cases.yaml --filter "$(FILTER)"; \
```
Widcket, line 132: Do we need `--as-machine` at all, if the user can simply pass the needed values through flags?

internal/cli/login.go
```diff
@@ -12,24 +12,68 @@ import (
...
 			cli.tracker.TrackCommandRun(cmd, cli.config.InstallID)
 
 			return nil
 		},
 	}
 
+	loginAsMachine.RegisterBool(cmd, &inputs.LoginAsMachine, false)
+	loginTenantDomain.RegisterString(cmd, &inputs.Domain, "")
+	tenantClientID.RegisterString(cmd, &inputs.ClientID, "")
+	tenantClientSecret.RegisterString(cmd, &inputs.ClientSecret, "")
```
Widcket, line 75: If `--client-id` and `--client-secret` cannot be used without `--as--machine`, maybe mark them as a [flag group](https://github.com/spf13/cobra/blob/main/user_guide.md#flag-groups) so cobra can enforce it?

Widcket: If we don't keep `--as-machine`, `--client-id` and `--client-secret` should probably still be marked as a group, as they need to be used together.

internal/cli/login.go
```diff
@@ -12,24 +12,68 @@ import (
...
+			cli.renderer.Infof("Successfully authenticated to %s", inputs.Domain)
 			cli.tracker.TrackCommandRun(cmd, cli.config.InstallID)
 
 			return nil
 		},
 	}
 
+	loginAsMachine.RegisterBool(cmd, &inputs.LoginAsMachine, false)
+	loginTenantDomain.RegisterString(cmd, &inputs.Domain, "")
+	tenantClientID.RegisterString(cmd, &inputs.ClientID, "")
```
Widcket, line 74: Since `tenants add` is being removed, `tenantClientID` and `tenantClientSecret` are no longer being used in the `tenants.go` file. Should we move their definitions to this file?

internal/cli/login.go
```diff
@@ -12,24 +12,68 @@ import (
...
 	cmd := &cobra.Command{
 		Use:   "login",
 		Args:  cobra.NoArgs,
 		Short: "Authenticate the Auth0 CLI",
-		Long:  "Sign in to your Auth0 account and authorize the CLI to access the Management API.",
+		Long:  "Authenticates the Auth0 CLI either as a user using personal credentials or as a machine using client credentials (client ID/secret).",
+		Example: `
+		auth0 login
+		auth0 login --as-machine
+		auth0 login --as-machine --domain <TENANT_DOMAIN> --client-id <CLIENT_ID> --client-secret <CLIENT_SECRET>
```
Widcket, line 48: Should we use lowercase for the placeholders? Like in other commands, e.g. https://github.com/auth0/auth0-cli/blob/main/internal/cli/apps.go#L256

internal/cli/login.go
willvedd: Using the flag group is a good suggestion, will implement.

internal/cli/login.go
```diff
@@ -12,24 +12,68 @@ import (
 	"github.com/auth0/auth0-cli/internal/prompt"
 )
 
+var (
+	loginAsMachine = Flag{
+		Name:       "Login as machine",
+		LongForm:   "as-machine",
+		Help:       "Initiates authentication as a machine via client credentials (client ID, client secret)",
+		IsRequired: false,
```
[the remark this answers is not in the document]
willvedd: Planning to remove this flag anyway.

internal/cli/login.go
willvedd: `--as-machine` flag has been removed so resolving this.

internal/cli/login.go
```diff
@@ -12,24 +12,68 @@ import (
...
+	}
+	loginTenantDomain = Flag{
+		Name:       "Tenant Domain",
+		LongForm:   "domain",
+		Help:       "Specifies tenant domain when authenticating via client credentials (client ID, client secret)",
+		IsRequired: false,
```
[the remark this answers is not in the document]
Widcket: Note that pretty much every non-global flag in the CLI has a short form, so removing them will touch nearly every command that takes flags.

sergiught (approved):

PR #547, opened 2022-12-06
DXCDT-267: Graceful handling of access token regeneration (2/x)

willvedd:
### 🔧 Changes

With the consolidation of `auth0 tenants add` into `auth0 login`, there are considerations to be made around access token regeneration and error handling when those tokens are unable to be regenerated.

This PR introduces advancements in token regeneration for both methods of authentication, including more fault-tolerance and better guidance to the end-user when things go wrong.

### 📚 References

Upstream PR: #546 

### 🔬 Testing

Manually tested by setting token expiry into the past, invalid client credentials, invalid refresh token.

### 📝 Checklist

- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [x] I have added documentation for all new/changed functionality (or N/A)

PR #551, opened 2022-12-08
DXCDT-298: Interactive login prompt (3/x)

willvedd:
### 🔧 Changes

Implements an initial interactive prompt when executing the `auth0 login` command. This will help navigate the user through the decision tree of authentication types, in order to reduce confusion. This prompt is optional and can be skipped either by passing the `--no-input` flag (ex: `auth0 login --no-input`) or by passing in the client credentials through flags.

This PR also provides content updates to existing prompts too in an effort to provide more clarity and guidance around the authentication process.

The requirements this intendeds to satisfy:
- Interactive navigation to any of the login mechanisms (user, client credentials, etc)
- Should be expandable to other types of login mechanisms (ex: auth token)
- Ability to "skip" to any login mechanisms using flags, ideally implicitly but otherwise explicitly
- When possible, inferring login mechanism from flags (ex: `--client-id` will skip to user login)

## Skipping

As User:
`$ auth0 login --no-input` or `$auth0 login --scopes`

Client Credentials:
`$ auth0 login --domain <DOMAIN> --client-id <CLIENT_ID> --client-secret <CLIENT_SECRET>`

Token:
`$ auth0 login --token <TOKEN>`

## Interactive Mode

**Initial Prompt**

```
How do you want to login?
 > User
 > Machine
 > Token
```

**User**

```
User -> Create URL -> Open Browser -> stuff
```

**Machine**

```
Machine
 > What is the domain?
 > What is the client ID?
 > What is the client secret?
```

**Token**

```
Token
 > What is the auth token value?
```

### 📚 References

Related PRs: 
- #546 
- #547 

### 📝 Checklist


internal/cli/login.go
```diff
@@ -14,34 +14,39 @@ import (
...
+		LongForm:   "client-secret",
+		Help:       "Client Secret of the application.",
+		IsRequired: false,
+	}
+
+	loginAsUser = Flag{
+		Name:       "Login as user",
+		LongForm:   "as-user",
+		Help:       "Initializes login as a user via device code flow.",
+		IsRequired: false,
```
willvedd, line 41: I know that we removed the `--as-machine` flag previously, but I did not know of another way of skipping the interactive prompt and proceeding straight to the user login mechanism. 

Alternatively, if this is not a use case that we wish to satisfy, we can always remove and force the user to interact with the initial prompt. But my hunch is that more inclined users would want a way to skip that. It may also help with certain quickstarts and tutorials.

willvedd: **Update:** @sergiught and I decided to nix the `--as-user` flag in favor of the `--no-input` flag because it already exists and satisfies that functionality requirement. Further, we don't expect users to be needing the "skip" mechanism. very often, so it is ok to slightly deprioritize.

sergiught (approved):

PR #554, opened 2022-12-13
DXCDT-297: Remove env var ingestion

willvedd:
### 🔧 Changes

Device code flow for private tenants will not be initially supported for v1. Because of this, we can simplify the code by remove instances of env var ingestion as well as the `envdecode` dependency. These values were previously available for modification if users were targeting a private cloud tenant, however that process has been deemed impractical. Instead, private cloud users are advised to authenticate via client credentials instead.

### 🔬 Testing

Will be tested during `auth0 login` in integration tests.

### 📝 Checklist


internal/cli/root.go
```diff
@@ -21,11 +21,6 @@ const rootShort = "Supercharge your development workflow."
 
 // Execute is the primary entrypoint of the CLI app.
 func Execute() {
-	// cfg contains tenant related information, e.g. `travel0-dev`,
```
willvedd, line 24: Removing this comment because it references a now irrelevant `AUTH0_API_KEY` env var and generally doesn't apply to this codebase anymore.

Widcket (approved):

PR #577, opened 2023-01-04
DXCDT-301: Refactor secret storage (1/2)

willvedd:
### 🔧 Changes

Eliminating three unnecessary data structures: `Keyring`, `TokenRetriever` and `SecretStore`. They were used to implement thin abstractions around the `go-keyring` library, but the abstraction themselves didn't offer any domain-specific value. Instead, this proposed refactor reorients secret management towards the domain. Specifically, introducing lighter functions like `StoreRefreshToken` and `GetRefreshToken` rather than composing the get/set operations manually in the code.

This modest refactor sets the foundation for more secrets to be stored in a similar way, particularly the client secret as seen in #578.

### 🔬 Testing

Updates to unit tests.

### 📝 Checklist


internal/auth/secrets.go
```diff
@@ -1,22 +1,32 @@
...
-	return keyring.Delete(namespace, key)
+func DeleteSecretsForTenant(tenant string) error {
+	var errs error
+
+	e := keyring.Delete(secretRefreshToken, tenant)
+	if e != nil {
+		errs = errors.Wrap(errs, e.Error())
+	}
+
+	return errs
```
willvedd, line 31: The refocused functions here are the bulk of this PR.

internal/auth/auth.go
```diff
@@ -19,8 +19,6 @@ import (
 const (
 	audiencePath           = "/api/v2/"
 	waitThresholdInSeconds = 3
-	// SecretsNamespace is the namespace used to set/get values from the keychain.
-	SecretsNamespace = "auth0-cli"
```
willvedd, line 23: This single namespace value is no longer sufficient because we will have multiple secret values and need to distinguish between the different types.

sergiught (approved):

Issue #575, opened 2022-12-26
Error when trying to edit login-id prompt text


daschi:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://auth0.github.io/auth0-cli/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-cli/releases/latest) of this project and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Hello! I am using the auth0 cli to update the prompt texts for `login-id`, however when I run:

`auth0 branding texts update -l en login-id`

I see the following error message:

```
!!     Uh oh. Something went wrong.
!!     If this problem keeps happening feel free to report an issue at
!!
!!     https://github.com/auth0/auth0-cli/issues/new/choose
```

If I run `auth0 branding texts update -l en login`

I am able to change the text for the login prompts, however I am using the identifier first login page so I believe I need to edit `login-id` texts.

Thank you for your help and let me know if I can provide more info!

### Expectation

Expect `auth0 branding texts update -l en login-id` to open an editor to edit the prompt texts.

### Reproduction

1. Given I am logged into my auth0 tenant
2. When I run `auth0 branding texts update -l en login-id`
3. I see a generic error that something went wrong.

### Auth0 CLI version

auth0 version 0.13.1 42a703dd07e148a3a74be787180ac1e9938b63cd

willvedd:
@daschi – Thanks for submitting this issue, it certainly appears to be frustrating. I'm currently in the middle of troubleshooting this and have determined that it's a duplicate of #515. For that reason, I'm opting to close this ticket and communicate all status updates in the other ticket. Thank you for your patience!

Issue #515, opened 2022-11-08
Always getting "something went wrong" error on `auth0 branding texts update signup` command


maggie9595:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://auth0.github.io/auth0-cli/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-cli/releases/latest) of this project and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

The same `auth0 branding texts update signup` command that we are running for our tenants in version `0.11.7` seems to be erroring out persistently after we updated to version `0.12.0`, with this error:
```
!!     Uh oh. Something went wrong.
!!     If this problem keeps happening feel free to report an issue at
!!
!!     https://github.com/auth0/auth0-cli/issues/new/choose
```

Reinstalling did not seem to fix the error either.

### Expectation

`auth0 branding texts update signup` should not error out in the new version of the CLI.

### Reproduction

- Update to newest version (0.12.0) of the CLI
- Try to run `auth0 branding texts update signup` command on any tenant
- Get persistent `something went wrong` error

### Auth0 CLI version

0.12.0

jdk2pq:
I'm having the same issues. Every prompt I've tried gives the above error except for `auth0 branding texts update login`.

maggie9595:
> I'm having the same issues with any `auth0 branding texts update` command. Every prompt I've tried gives the above error, except for `auth0 branding texts update login`.

This is actually the case for me as well, just updated my issue!

jameskyker:
Auth0's API is broken. We tried to issue the GETs [described here](https://auth0.com/docs/api/management/v2#!/Prompts/get_custom_text_by_language), and none of them seem to work except for login. The login prompt returns values, but signup, for instance, returns an empty object as well as every other we tried. Auth0 needs to fix their API, and then this client will probably work.

jameskyker:
[...]

PR #584, opened 2023-01-05
DXCDT-302: Fix prompts update failing for non-login prompt types

willvedd:
### 🔧 Changes

As reported by #515, #575, the command `auth0 branding texts update <PROMPT_TYPE>` will always fail when the prompt type is not `login`. For example, `auth0 branding texts update signup` would always fail. 

This failure is caused by faulty logic that doesn't adequate traverse a set of default values for the provide prompt type, but this is better illustrated in the diff below. Further, if the prompt type is not found in the set of default values, which could happen if the default values change or new prompt types are added, the editor will start with an empty object instead of panicking.

### 📚 References

Related Issues: #515, #575

### 🔬 Testing

Added two integration tests with fixtures to show that multiple prompt types work now.

### 📝 Checklist


internal/cli/prompts_custom_text.go
```diff
@@ -214,9 +214,12 @@ func downloadDefaultBrandingTextTranslations(prompt, language string) map[string
 			return nil
 		}
 
-		selectedPrompt := allPrompts[0][prompt].(map[string]interface{})
```
willvedd, line 217: The issue is that the `allPrompts` is a set of twenty-something values, each with a single-property object pertaining to a specific prompt type.

This is why the `login` prompt was working, because it was the zeroth index in this set.

sergiught (approved):

Widcket (approved):

PR #677, opened 2023-03-16
DXCDT-413: `utils_shared.go` 

willvedd:
### 🔧 Changes

Adding a few basic tests to the `utils_shared.go` file to get some quick coverage wins. Unable to test remaining functions in file without some changes to the abstractions to make them more testable.

### 🔬 Testing

This _only_ adds testing.

### 📝 Checklist


internal/cli/utils_shared_test.go
```diff
@@ -0,0 +1,59 @@
...
+	assert.True(t, containsStr([]string{"string-1", "string-2"}, "string-1"))
+}
+
+func TestGenerateState(t *testing.T) {
+	state, err := generateState(0)
+	assert.Equal(t, "", state)
+	assert.Nil(t, err)
+
+	state, err = generateState(cliLoginTestingStateSize)
+	assert.IsType(t, "string", state)
```
Widcket, line 57:
```suggestion
	assert.IsType(t, "string", state)
	assert.Equal(t, cliLoginTestingStateSize, len(state))
```
I'd suggest adding this assert (I haven't tried it out though).

willvedd: Turns out this won't work because `generateState` base64 encodes the string, which will change the expected length of the string.

ewanharris (approved):

PR #688, opened 2023-03-29
Combining test scripts for predictable and isolate test cases

willvedd:
### 🔧 Changes

In order for many integration tests to pass with strong assertions, test resources must exist on the testing tenant. Previously, this was a two step process, creating a resource and then fetching the ID. However, this meant that in order to use a test resource's ID in a test, it needs to be explicitly created beforehand. With many Auth0 resources existing in a web of co-dependencies, it is arduous to orchestrate the creation of these test resources and prohibits test files from being run in isolation. Instead, this PR combine the process into a single script, precluding the need to explicitly define a steps for creating test resources.

The result are tests that are simpler, eliminate dependencies and increase the ability to run tests in isolation.

### 🔬 Testing

Manually tested to confirm that tests run in isolation. All tests should pass here.

### 📝 Checklist

### DIFF
--- a/test/integration/scripts/get-api-id.sh
+++ b/test/integration/scripts/get-api-id.sh
@@ -1,6 +1,13 @@
 #! /bin/bash
 
+FILE=./test/integration/identifiers/api-id
+if [ -f "$FILE" ]; then
+    cat $FILE
+    exit 0
+fi
+
 api=$( auth0 apis create --name integration-test-api-newapi --identifier http://integration-test-api-newapi --scopes read:todos --json --no-input )
 
 mkdir -p ./test/integration/identifiers
-echo "$api" | jq -r '.["id"]' > ./test/integration/identifiers/api-id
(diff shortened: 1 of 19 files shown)


test/integration/scripts/get-api-id.sh
```diff
@@ -1,6 +1,13 @@
...
+    cat $FILE
+    exit 0
+fi
+
 api=$( auth0 apis create --name integration-test-api-newapi --identifier http://integration-test-api-newapi --scopes read:todos --json --no-input )
 
 mkdir -p ./test/integration/identifiers
-echo "$api" | jq -r '.["id"]' > ./test/integration/identifiers/api-id
+echo "$api" | jq -r '.["id"]' > $FILE
+cat $FILE
```
willvedd, line 13: This change to the `get-api-id` script is the same for all other scripts and represents the core of the changes in this PR.

ewanharris (approved):

PR #690, opened 2023-03-29
DXCDT-414: Custom domains integration tests

willvedd:
### 🔧 Changes

Adding integration tests to the suite of custom domains commands. 

### 🔬 Testing

Only testing changes made.

### 📝 Checklist

### DIFF
--- a/test/integration/custom-domains-test-cases.yaml
+++ b/test/integration/custom-domains-test-cases.yaml
@@ -1,0 +1,74 @@
+config:
+  inherit-env: true
+tests:
+  001 - list domains:
+    command: auth0 domains list
+    exit-code: 0
+  002 - create domain with minimal flags:
+    command: auth0 domains create --domain "integration-test.com" --no-input
+    exit-code: 0
+    stdout:
+      contains:
+        - "ID                       cd_"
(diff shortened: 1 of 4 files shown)


test/integration/custom-domains-test-cases.yaml
```diff
@@ -0,0 +1,74 @@
...
+    command: auth0 domains update $(./test/integration/scripts/get-custom-domain-id.sh) --policy recommended --no-input
+    exit-code: 0
+    stdout:
+      contains:
+        - "ID                       cd_"
+        - "DOMAIN                   integration-test.com"
+        - "STATUS                   pending_verification"
+        - "PROVISIONING TYPE        auth0_managed_certs"
+        - "TLS POLICY               recommended"
+  007 - verify domain:
```
ewanharris, line 50: Just a nit query here about wording as the assertion still says `pending_verification` vs the test case being `verify domain`. Other than that, looks good

willvedd: Agreed. Removed because haven't set up a custom domain yet and if we do, this assertion could break. Just being able to call the management API verify endpoint is enough for me IMO.

ewanharris (approved):

PR #691, opened 2023-03-30
Add test for deleting all tenant secrets from keyring

ewanharris:
### 🔧 Changes

This adds a test that stores tokens/secrets, calls `DeleteSecretsForTenant` to delete them, and then asserts they were deleted.

Also fixes an issue where the secret not found error was not being returned for GetAccessToken as looking it up would return an empty string and nil on the first token lookup as it satisfied the `err == keyring.ErrNotFound` condition. The change here is to only return the accessToken early if we have pulled more than 1 item from the keyring 

### 🔬 Testing

Adds tests mostly, and change included is covered by said tests

### 📝 Checklist

### DIFF
--- a/internal/keyring/keyring.go
+++ b/internal/keyring/keyring.go
@@ -88,7 +88,9 @@ func GetAccessToken(tenant string) (string, error) {
 
 	for i := 0; i < secretAccessTokenMaxChunks; i++ {
 		a, err := keyring.Get(fmt.Sprintf("%s %d", secretAccessToken, i), tenant)
-		if err == keyring.ErrNotFound {
+		// Only return if we have pulled more than 1 item from the keyring, otherwise this will be
+		// a valid "secret not found in keyring"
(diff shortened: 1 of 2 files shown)


internal/keyring/keyring.go
```diff
@@ -88,7 +88,9 @@ func GetAccessToken(tenant string) (string, error) {
 
 	for i := 0; i < secretAccessTokenMaxChunks; i++ {
 		a, err := keyring.Get(fmt.Sprintf("%s %d", secretAccessToken, i), tenant)
-		if err == keyring.ErrNotFound {
+		// Only return if we have pulled more than 1 item from the keyring, otherwise this will be
+		// a valid "secret not found in keyring"
+		if err == keyring.ErrNotFound && i > 1 {
```
willvedd, line 93:
```suggestion
		if err == keyring.ErrNotFound && i > 0 {
```

ewanharris: D'oh, of course!

PR #694, opened 2023-03-30
DXCDT-417:  Tests for `test` commands

willvedd:
### 🔧 Changes

Adding integration tests for both the `auth0 test token` and `auth0 test login` commands. The bulk of this PR is orchestration of clients and client grants through scripts to emulate certain scenarios. 

### 🔬 Testing

Almost only testing additions.

### 📝 Checklist


internal/cli/test.go
```diff
@@ -269,7 +269,6 @@ func testTokenCmd(cli *cli) *cobra.Command {
 	cmd.SetUsageTemplate(resourceUsageTemplate())
 	cmd.Flags().BoolVar(&cli.force, "force", false, "Skip confirmation.")
 	cmd.Flags().BoolVar(&cli.json, "json", false, "Output in json format.")
-	testAudience.IsRequired = true
```
willvedd, line 272: Requiring the audience precluded the ability to test authentication in non-interactive mode.

test/integration/test-commands-test-cases.yaml
```diff
@@ -0,0 +1,38 @@
+config:
+  inherit-env: true
+
+tests:
+  001 - test login of default app (partially successful):
+    #Terminate early because command will otherwise wait for in-browser authorization stage
+    command: timeout 2s  auth0 test login $(./test/integration/scripts/get-default-app-id.sh) --no-input 2>&1
```
willvedd, line 7: Admittedly, looks odd. What's happening is running the `auth0 test login` command but enforcing a 2s timeout and piping stderr into stdout. Otherwise, it will wait indefinitely for the user to authorize in the broswer. By terminating early, this test only verifies the first stage, but its better than nothing.

internal/cli/test.go
ewanharris: It looks like the initial change of toggling this to be required is in #221 otherwise it generated invalid tokens. Testing locally it still seems to be that dropping audience will generate an invalid token due to a missing payload.

test/integration/test-commands-test-cases.yaml
ewanharris: Is `timeout` a default command on MacOS? I don't appear to have it, maybe I need to install coreutils or something (I think that's fine though, maybe just needs a contributing note?)

internal/cli/test.go
willvedd: Let's talk about this later.

willvedd: **Update:** Turns out that this marks the audience flag required for both the `test token` and `test login` commands but it should only be requiring for `test token` command. Will be fixing in a following commit.

ewanharris (approved):

PR #695, opened 2023-03-31
Add test for auth0 role permissions

ewanharris:
### 🔧 Changes

Adds tests that cover the `auth0 role permissions` command. I wasn't sure whether this should be in a `role-permissions-test-cases.yaml` or `role-test-cases.yaml` so happy to move depending on feedback

### 🔬 Testing

This solely adds tests

### 📝 Checklist

### DIFF
--- a/test/integration/roles-permissions-test-cases.yaml
+++ b/test/integration/roles-permissions-test-cases.yaml
@@ -1,0 +1,49 @@
+config:
+  inherit-env: true
+
+tests:
+  001 - list role permissions:
+    command: auth0 roles permissions ls $(./test/integration/scripts/get-role-id.sh)
+    exit-code: 0
+    stderr:
+      contains:
+        - No role permissions available.
+
+  002 - list role permissions:
+    command: auth0 roles permissions ls $(./test/integration/scripts/get-role-id.sh) --number -1
+    exit-code: 1
+    stderr:
+      contains:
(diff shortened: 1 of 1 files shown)


test/integration/roles-permissions-test-cases.yaml
```diff
@@ -0,0 +1,49 @@
+config:
+  inherit-env: true
+
+tests:
+  001 - list role permissions:
+    command: auth0 roles permissions ls $(./test/integration/scripts/get-role-id.sh)
```
Widcket, line 6: Should we add a test where we assign a permission before hand (via script) and the assert the output? We can keep this one as well as it asserts the empty state.

Widcket: My bad, I just saw test `004` 🤦🏼‍♀️

test/integration/roles-permissions-test-cases.yaml
```diff
@@ -0,0 +1,49 @@
+config:
+  inherit-env: true
+
+tests:
+  001 - list role permissions:
```
Widcket, line 5:
```suggestion
  001 - list role permissions with empty state:
```

Widcket: Or `with no permissions`.

PR #710, opened 2023-04-03
Add missing integration tests

Widcket:
### 🔧 Changes

This PR adds a series of missing integration tests and also updates existing ones, exercising missing commands, flags, and assertions.

Also the custom domain `integration-test.com` was changed to `custom-domain.com` to work around an issue with the integrations tests tenant, as informed in Slack. 

### 🔬 Testing

This PR just adds/updates integration tests.

### 📝 Checklist

- [X] All new/changed/fixed functionality is covered by tests (or N/A)
- [X] I have added documentation for all new/changed functionality (or N/A)


internal/cli/root.go
```diff
@@ -44,9 +44,8 @@ func Execute() {
 		if v := recover(); v != nil {
 			err := fmt.Errorf("panic: %v", v)
 
-			// If we're in development mode, we should throw the
-			// panic for so we have less surprises. For
-			// non-developers, we'll swallow the panics.
+			// If we're in development mode, we should throw the panic for so we have less surprises.
+			// For non-developers, we'll swallow the panics.
```
Widcket, line 48: This comment is not excluded by Codecov: https://app.codecov.io/gh/auth0/auth0-cli/blob/main/internal/cli/root.go#L47

internal/cli/root.go
```diff
@@ -44,9 +44,8 @@ func Execute() {
...
 			if instrumentation.ReportException(err) {
 				fmt.Print(panicMessage)
 			} else {
```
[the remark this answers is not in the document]
willvedd: Maybe we can do something like this to shed a few extra lines? Not this exact suggestion, it only just serves as an example.

internal/cli/root.go
```diff
@@ -44,12 +44,12 @@ func Execute() {
 		if v := recover(); v != nil {
 			err := fmt.Errorf("panic: %v", v)
 
-			// If we're in development mode, we should throw the
-			// panic for so we have less surprises. For
-			// non-developers, we'll swallow the panics.
+			// If we're in development mode, we should throw the panic for so we have less surprises.
+			// For non-developers, we'll swallow the panics.
```
Widcket, line 48:
```suggestion
```

test/integration/test-cases.yaml
```diff
@@ -202,3 +251,17 @@ tests:
...
+    stderr:
+      contains: 
+        - "Default tenant switched to:"
+
+  tenants open:
+    command: auth0 tenants open $AUTH0_CLI_CLIENT_DOMAIN --no-input
+    exit-code: 0
+    stderr:
+      contains: 
+        - "Open the following URL in a browser:"
```
willvedd, line 267: I think it would be nice to have a soft assertion that an actual URL is printed out. Maybe something like:

```suggestion
        - "Open the following URL in a browser: https://manage.auth0.com/dashboard/"
```

willvedd (approved):

PR #917, opened 2023-11-20
DXCDT-597: Don't require all scopes for client credentials auth

willvedd:
### 🔧 Changes

As noted by #556, when authenticating to the CLI with client credentials, it is necessary for that client to have a set of required scopes. Doing so ensures that the CLI has full functionality out-the-box, however, this precludes the CLI from adhering to principle of least privilege. As the original Github issue notes, the CLI can be used in automation for specific cases and users should be able to provision only the scopes for those cases.

### 📚 References

Related Github issue: #556 

### 🔬 Testing

Manual verification; no testing framework in place to test token exchange.

### 📝 Checklist


internal/auth/auth.go
```diff
@@ -116,6 +116,29 @@ func WaitUntilUserLogsIn(ctx context.Context, httpClient *http.Client, state Sta
 	}
 }
 
+var RequiredScopes = []string{
```
willvedd, line 119: IMO, doesn't make sense to have these scopes as a standalone file. Decided to pull-in to the `auth.go` file.

sergiught (approved):

willvedd:
Confirmed again that this works as expected; no other logic needs to be modified.

Issue #804, opened 2023-08-18
auth0 users blocks list <user-id> and auth0 users blocks unblock <user-id> should allow name and email as indefier


msmolka:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://auth0.github.io/auth0-cli/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-cli/releases/latest) of this project and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Unblock user should be possible also via username and user email. currently only user id is possible.
management API uses additional identifiers:

Get blocks by identifier

Retrieve a list of blocked IP addresses for a given identifier (e.g., username, phone number or email).
Endpoint
GET  /api/v2/user-blocks

### Expectation

the cli is consitent with management api

### Reproduction

give command `auth0 users blocks list some@example.com`
should return blocks for given user
but returning error:
`failed to list user blocks for user with ID some@example.com: 400 Bad Request: Path validation error: 'Object didn't pass validation for format user-id: kirill.owen@bigchange.com' on property id (user_id of the user blocks to retrieve).`

### Auth0 CLI version

1.0.1

willvedd:
Appreciate the feedback, I agree that you _should_ be able to block and unblock users using alternative identifiers. Marking this as a feature request for now.

PR #931, opened 2023-12-01
DXCDT-601: Supporting multiple identifiers for blocks list and unblock commands

willvedd:
### 🔧 Changes

Currently, the `auth0 users blocks list` and `auth0 users blocks unblock` commands only support user ID as the identifier. Additional identifiers were requested in #804. To address this request, this PR adds the support for email, username and phone number as identifiers for these commands in addition to the user ID. 

This 

### 📚 References

Related Issue: #804 

### 🔬 Testing

Adding integration test cases.

### 📝 Checklist


internal/cli/users_blocks.go
```diff
@@ -25,32 +26,39 @@ func userBlocksCmd(cli *cli) *cobra.Command {
...
 			var userBlocks []*management.UserBlock
 			err := ansi.Waiting(func() (err error) {
-				userBlocks, err = cli.api.User.Blocks(cmd.Context(), inputs.userID)
+				userBlocks, err = cli.api.User.Blocks(cmd.Context(), inputs.userIdentifier)
+				if mErr, ok := err.(management.Error); ok && mErr.Status() != http.StatusBadRequest {
+					return nil
+				}
+
+				userBlocks, err = cli.api.User.BlocksByIdentifier(cmd.Context(), inputs.userIdentifier)
 				return err
```
m3talsmith, line 58: This looks like it implicitly returns one of the possible two errors: it could be the first error, the second error, or nil. Should we check for the first error and shortcut the process if it exists?

internal/cli/users_blocks.go
```diff
@@ -65,29 +73,37 @@ func listUserBlocksCmd(cli *cli) *cobra.Command {
...
 					return err
 				}
 			} else {
-				inputs.userID = args[0]
+				inputs.userIdentifier = args[0]
 			}
 
 			err := ansi.Spinner("Unblocking user...", func() error {
-				return cli.api.User.Unblock(cmd.Context(), inputs.userID)
+				err := cli.api.User.Unblock(cmd.Context(), inputs.userIdentifier)
```
m3talsmith, line 98: There's a possible dangling error here as well.

m3talsmith (commented):
The only possible changes that I can see here are maybe adding some error catching (as commented on the lines).

m3talsmith (approved):

Issue #414, opened 2022-04-08
Add metadata option on `apps update`


amadeu01:
### Describe the problem you'd like to have solved

I would like to be able to update application metadata info.

<img width="955" alt="image" src="https://user-images.githubusercontent.com/11631092/162451895-73b2be5b-6df0-48a0-8b59-5bf82c2a0360.png">

### Describe the ideal solution

The ideal solution would he an option as other updates

```
auth0 apps update <id> --metadata "key" "value"
```

ramya18101:
PR got approved and closing this issue

PR #938, opened 2023-12-07
DXCDT:596 Add Metadata Flag to Apps

m3talsmith:
### 🔧 Changes

Added flag functionality to add client metatdata to apps create and update.

### 📚 References

This addresses issue #414 , the main difference being the format of the metadata when passed.

The previous ticket suggested key / value definition as:

`--metadata "key" "value"`

This may work if it was an array, but Metadata is a map, hence this key / value definition:

`--metadata "key=value"`

Multiple of the flag or comma delimited values will concatenate into new map key / values. Such that:

- `--metadata "key=value" --metadata "foo=bar"`
- `--metadata "key=value,foo=bar"`

Create the same map: `map[key:value foo:bar]`.

### 🔬 Testing

1. `auth0 apps create -n mymetaapp -d mymetaapp -t m2m -r --json --metadata "foo=bar"`
2. View your app metadata on the web portal, in advanced settings
3. `auth0 apps update <app-id> -n mymetaapp -d mymetaapp -t m2m -r --json --metadata "baz=buzz"`
4. View your updated app metadata on the web portal, in advanced settings

### 📝 Checklist


internal/cli/apps.go
```diff
@@ -638,8 +659,13 @@ func updateAppCmd(cli *cli) *cobra.Command {
...
 			}
 
+			clientMetadata := make(map[string]interface{}, len(inputs.Metadata))
+			for k, v := range inputs.Metadata {
+				clientMetadata[k] = v
+			}
+
 			// Load updated values into a fresh app instance
-			a := &management.Client{}
+			a := &management.Client{ClientMetadata: &clientMetadata}
```
willvedd, line 668: One thing to verify here is if a user doesn't specify metadata when updating, effectively skipping it, does that overwrite any previously existing metadata on remote?

internal/cli/apps.go
```diff
@@ -78,6 +78,12 @@ var (
 		IsRequired:   false,
 		AlwaysPrompt: true,
 	}
+	appMetadata = Flag{
+		Name:       "Metadata",
+		LongForm:   "metadata",
+		Help:       "Will add Metadata to an application",
```
willvedd, line 84:
```suggestion
		Help:       "Arbitrary keys-value pairs (max 255 characters each), that  can be assigned to each application. More about application metadata: https://auth0.com/docs/get-started/applications/configure-application-metadata",
```

willvedd (commented):
A couple small callouts. Also, remember to update the commander integration tests. The existing application commands could be updated and potentially a new case made specifically for metadata.

internal/cli/apps.go
m3talsmith: Changed appropriately. Great catch!

willvedd (approved):
👍

Issue #936, opened 2023-12-06
logs ls --number has no effect


jonaustin:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://auth0.github.io/auth0-cli/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-cli/releases/latest) of this project and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

```
❯ auth0 logs ls --number 100000 |wc -l

=== xxxx.us.auth0.com logs

     101
❯ auth0 logs ls |wc -l

=== xxx.us.auth0.com logs

     101
```

Poked at the code, and there's simply no code that loops over pages:
https://github.com/auth0/auth0-cli/blob/main/internal/cli/logs.go#L172-L189

### Expectation

Passing in -n/--number >100 should fetch more than 100 log entries.

### Reproduction

auth0 logs ls --number 100000

### Auth0 CLI version

❯ auth0 -v
auth0 version 1.3.0 2bd069cfc1ee0501ed039fc80b2e550447c8ef14

willvedd:
This is fixed in #939 but still waiting on a release.

jonaustin:
@willvedd Why is the maximum only 1000?

That greatly reduces the utility of this in my mind; i.e. it's still not really all that useful if it can only go back 1000 log entries.

willvedd:
Arbitrarily set at 1000 for practically; there is no technical limitation. IMO, this will satisfy most (but obviously not all) use cases around logs. And we need to bound at some point. 

It would help to understand the workflow you're trying to achieve. Short of upping the limit a little, we're unlikely to immediately implement but it is still important for us to know the features you want for the CLI.

DannyPat44:
I would also like to see a higher number, but primarily so that I can determine easily how many particular errors of a specific kind have occurred. If a utility for the CLI allowed you to do this, it would probably be okay to cap the total number returned at 1000. 

[...]

PR #939, opened 2023-12-07
Adding request pagination to `auth0 logs list`

willvedd:
### 🔧 Changes

As noted by #936 , the `auth0 logs list` command's `--number` flag does not have an effect for any value above 100. This PR introduces a pagination mechanism such that the correct number of logs are requested. 

### 📚 References

Related issue: #936 

### 🔬 Testing

Added integration test cases .

However, the test tenant won't reliably have a high number of logs, which makes testing the pagination functionality tricky. For these cases, I tested locally against my tenant with `--number` equaling 1,499,500,501 and 1000.

### 📝 Checklist

### DIFF
--- a/docs/auth0_logs_list.md
+++ b/docs/auth0_logs_list.md
@@ -22,7 +22,7 @@ auth0 logs list [flags]
   auth0 logs list --filter "user_name:<user-name>"
   auth0 logs list --filter "ip:<ip>"
   auth0 logs list --filter "type:f" # See the full list of type codes at https://auth0.com/docs/logs/log-event-type-codes
-  auth0 logs ls -n 100
+  auth0 logs ls -n 250
 ```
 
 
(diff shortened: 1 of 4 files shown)


docs/auth0_logs_tail.md
```diff
@@ -22,15 +22,15 @@ auth0 logs tail [flags]
   auth0 logs tail --filter "user_name:<user-name>"
   auth0 logs tail --filter "ip:<ip>"
   auth0 logs tail --filter "type:f" # See the full list of type codes at https://auth0.com/docs/logs/log-event-type-codes
-  auth0 logs tail -n 100
+  auth0 logs tail -n 10
```
willvedd, line 25: When tail'ing, its not as useful to initially print out a large number of requests. The output will continue to append, so its not as if this is the effective size of display. In fact, I doubt most users would want to tail out a particular number of logs.

m3talsmith (approved):

docs/auth0_logs_list.md
```diff
@@ -31,7 +31,7 @@ auth0 logs list [flags]
 ```
   -f, --filter string   Filter in Lucene query syntax. See https://auth0.com/docs/logs/log-search-query-syntax for more details.
       --json            Output in json format.
-  -n, --number int      Number of log entries to show. (default 100)
+  -n, --number int      Number of log entries to show. Minimum 1, maximum 1000. (default 100)
```
jonaustin, line 34: Why is the maximum only 1000?

That greatly reduces the utility of this in my mind.

PR #941, opened 2023-12-11
Manual installation and environment variables

andychiare:
### 🔧 Changes

I installed the Auth0 CLI on my Windows machine following the [manual installation instructions](https://github.com/auth0/auth0-cli?tab=readme-ov-file#manual) because I don't want to install Scoop (by the way, as far as I know, using Scoop is unusual for Windows users, an MSI package would be more idiomatic).

However, to be fully functional, you need to:
- add the folder where you extracted the binary to the `PATH` environment variable. This allows the `auth0` command to be accessible everywhere.
- create a `HOME` environment variable and set its value to the folder where you extracted the binary. This allows the `auth0` executable to find the `.config` file and prevents the user from getting an error.

I suppose that this applies to Linux and macOS as well, but not sure.

### 📝 Checklist


README.md
```diff
@@ -79,7 +79,8 @@ go install github.com/auth0/auth0-cli/cmd/auth0@latest
    - **macOS**: `$ tar -xf auth0-cli_{version}_Darwin_{architecture}.tar.gz`
    - **Linux**: `$ tar -xf auth0-cli_{version}_Linux_{architecture}.tar.gz`
    - **Windows**: Extract `auth0-cli_{version}_Windows_{architecture}.zip` using your preferred method of choice
-3. Run `./auth0`
+3. Make sure that the `PATH` and `HOME` environment variables include the folder where the binary was extracted.
+4. Run `./auth0`
```
m3talsmith, line 83: Maybe we should remove the `./` in order to reflect the cli being in the `PATH` var now.

m3talsmith: @andychiare do you mind making this change?

andychiare: Done! @m3talsmith

m3talsmith (approved):

PR #946, opened 2023-12-16
Update Windows installation instructions

deepu105:
Add instructions to install via Powershell commands

### 📝 Checklist


README.md
```diff
@@ -49,6 +49,17 @@ scoop bucket add auth0 https://github.com/auth0/scoop-auth0-cli.git
 scoop install auth0
 ```
 
+Install via [Powershell](https://learn.microsoft.com/en-us/powershell/)
+
+```powershell
+# Download the binary to the current folder. Use the desired version number.
+$version = "1.2.0"
```
willvedd, line 56: Is there a way to pull in the latest version instead of hardcoding? 1.2.0 is already out-of-date and I'd prefer not need to update this value whenever we make a release.

deepu105: Maybe we can make this into a powershell script, like the one for linux/mac so user just needs to execute it

m3talsmith: @deepu105 I like the PowerShell script approach as it matches the Bash script approach. Let's head in that direction.

developerkunal:
Hi @deepu105,

I hope you're having a nice day!

I've made some changes that will fetch the latest version for the tool. Could you please add those changes to this PR: [patch-1](https://github.com/auth0/auth0-cli/tree/patch-1)?

Thank you!

deepu105:
closing in favor of https://github.com/auth0/auth0-cli/pull/1020 cc @developerkunal

PR #1038, opened 2024-07-15
Added new login flow. Respect domain flag on auth0 login command. 

duedares-rvj:
New improved login flow.
When domain flag is passed along with --no-input during auth0 login,  it is interpreted as a user-flow. And hence pre-set the tenant while performing the browser based login. 

The evaluation matrix of the flow is mentioned here:

Screenshot 2024-08-05 at 1 14 30 PM

### 🔧 Changes

1. Added new flow to determine user-flow or machine-flow based on the initial cli input. (Handle params like --no-input)
2. If the evaluation cannot be made, prompt user for login type.
3. Added domain value evaluation function ensuring it's auth.com domain.
4. Relevant unit test cases added.

Eg. User could use below for user-login:
```auth0 login --domain <tenant domain> --no-input```

### 📚 References

This addresses a feature request as suggest in: 
https://github.com/auth0/auth0-cli/issues/1000

### 🔬 Testing
Required test cases added under login_test.go which is consumed by make commands. No additional steps required.

### 📝 Checklist

### CODE REVIEW

File: internal/auth/auth.go [hunk 1]
```diff
@@ -144,13 +144,13 @@ var RequiredScopes = []string{
 // GetDeviceCode kicks-off the device authentication flow by requesting
 // a device code from Auth0. The returned state contains the
 // URI for the next step of the flow.
-func GetDeviceCode(ctx context.Context, httpClient *http.Client, additionalScopes []string) (State, error) {
+func GetDeviceCode(ctx context.Context, httpClient *http.Client, additionalScopes []string, domain string) (State, error) {
```
acwest, line 147: You need to update the tests in `auth_test.go`

duedares-rvj: Added the new commit. Thank you!

internal/cli/login_test.go
```diff
@@ -0,0 +1,86 @@
...
+		cli := &cli{
+			renderer: &display.Renderer{
+				MessageWriter: message,
+				ResultWriter:  result,
+			},
+			noInput: true,
+		}
+
+		cmd := loginCmd(cli)
+		cmd.SetArgs([]string{"--client-id", "t3dbMFeTokYBguVu1Ty88gqntUXELSn9", "--client-secret", "3OAzE7j2HTnGOPeCRFX3Hg-0sipaEnodzQK8xpwsRiTkqdjjwEFT04rgCp4PZL4Z", "--domain", "duedares.us.auth0.com"})
```
duedares-rvj, line 80: @developerkunal @acwest How does it work around here?
For positive test case, should I store client-secret/id in plain text here?

developerkunal:
Changes LGTM.

internal/cli/login_test.go
acwest: I think we might have to add go-vcr to this, I can't see any good way to handle the credentials in the test

acwest: The alternative (which I believe we do in most places) the credentials get passed in the environment, and that can be set up in secrets for CI

acwest (approved):

PR #1150, opened 2025-03-07
Handle nil check on ReadRendering management API Response

ramya18101:
- Added nil checks to the ReadRendering management API response before unmarshalling to prevent potential panic errors

### DIFF
--- a/internal/cli/universal_login_customize.go
+++ b/internal/cli/universal_login_customize.go
@@ -409,7 +409,7 @@ func fetchRenderSettings(cmd *cobra.Command, cli *cli, input customizationInputs
 		return nil, fmt.Errorf("failed to fetch the existing render settings: %w", err)
 	}
 
-	// Marshal existing render settings into JSON and parse into a map if it's not nil
+	// Marshal existing render settings into JSON and parse into a map if it's not nil.
 	if existingRenderSettings != nil {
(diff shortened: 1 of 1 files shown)


duedares-rvj (approved):

internal/cli/universal_login_customize.go
```diff
@@ -322,37 +322,36 @@ func customizeUniversalLoginCmd(cli *cli) *cobra.Command {
...
 	if err != nil {
 		return err
 	}
 
 	renderSettings, err := fetchRenderSettings(cmd, cli, input)
-	if errors.Is(err, ErrNoChangesDetected) {
-		if renderSettings != nil && renderSettings.RenderingMode != nil {
-			currMode = string(*renderSettings.RenderingMode)
-		}
+	if renderSettings != nil && renderSettings.RenderingMode != nil {
```
developerkunal, line 333: shouldn't we check error first and then check the renderSettings

ramya18101: Usually Yes, But in this scenario, we need to callout the rendering Mode of UL if the error is `ErrNoChangesDetected`. So, need to fetch the renderingMode from the renderSettings before checking on error..!

developerkunal (approved):
LGTM

PR #1391, opened 2025-11-26
feat: add interactive api selection for resource server client creation

patrickkang:
### 🔧 Changes

Improves the DX for creating resource server clients by adding an interactive API picker instead of requiring manual entry of the resource server identifier:
- When creating a resource server client without the `--resource-server-identifier` flag, users are now prompted to select from their existing APIs via an interactive picker
- Added validation to reject empty or whitespace-only identifiers when provided via flag
- Non-interactive usage with `--resource-server-identifier` flag continues to work as before

### 📚 References
- https://auth0team.atlassian.net/browse/AIDX-332

### 🔬 Testing

```
# prompts for selection
auth0 apps create --name "My Client" --type resource_server

# non-interactive(skips prompt)
auth0 apps create --name "My Client" --type resource_server --resource-server-identifier "https://api.example.com"
```

image

### 📝 Checklist

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -515,7 +515,7 @@ func createAppCmd(cli *cli) *cobra.Command {
 			if appIsResourceServer {
 				if !appResourceServerIdentifier.IsSet(cmd) {
 					var selectedAPIID string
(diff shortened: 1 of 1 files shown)


ramya18101:
@patrickkang, Thanks for raising the PR..
We will review and get back to you soon!

PR #1476, opened 2026-03-30
feat: project detection support added

KartikJha:
#### `internal/cli/quickstart_detect.go` _(new file)_
- Added `DetectProject(dir string) DetectionResult` — scans the working directory for framework signal files and returns a structured result with `Framework`, `Type`, `BuildTool`, `Port`, `AppName`, `Detected`, and `AmbiguousCandidates`
- Implements 14 detection priority levels covering: Angular, Flutter (web/native), Vite+React/Vue/Svelte/Vanilla, Next.js, Nuxt, SvelteKit, Expo, ASP.NET (.csproj), Java (Spring Boot / Java EE / Vanilla via pom.xml or build.gradle), PHP/Laravel (composer.json), Go (go.mod), Ruby on Rails (Gemfile), Python/Flask (requirements.txt / pyproject.toml), and package.json dep scanning for Express/Fastify/Hono/Ionic/React Native
- Config-file signals (e.g. `angular.json`, `vite.config.ts`) take priority over generic dep scanning to avoid false positives
- When multiple package.json deps match, `AmbiguousCandidates` is populated so the caller can prompt for disambiguation

#### `internal/cli/quickstarts.go`

**New `setup-experimental` command flow:**
- Added `SetupInputs` named struct replacing 3x repeated anonymous structs
- Added `resolveRequestParams()` — resolves `DetectionSub` placeholders in `RequestParams` (callbacks, logout URLs, web origins, name) using actual app name and port at runtime
- Integrated `DetectProject()` into `RunE`: on an existing project the detected framework/type/build-tool/port/name are shown as a summary and the user is asked to confirm; accepted values pre-fill `SetupInputs` so downstream prompts are skipped
- Ambiguous detection (multiple package.json deps) triggers a framework disambiguation prompt
- No-detection fallback (empty directory) prompts for all inputs: type -> framework (filtered by type) -> build tool -> port -> name
- `getQuickstartConfigKey` skips any field already populated (by flags or detection) and only asks for what is missing
- API creation now uses `management.ResourceServer` + `cli.api.ResourceServer.Create()` instead of `management.Client`
- API name defaults to `<app-name>-API`; API identifier defaults to `https://<app-name-slug>`
- API-only flow prompts the user to select an existing Auth0 application to associate with
- Registered all spec flags: `--app`, `--api`, `--name`, `--type`, `--framework`, `--build-tool`, `--port`, `--callback-url`, `--logout-url`, `--web-origin-url`, `--identifier`, `--audience`, `--signing-alg`, `--scopes`, `--token-lifetime`, `--offline-access`

**Other fixes applied during review:**
- Removed hardcoded `RS256` overwrite on API signing algorithm
- Secret fields (e.g. `AUTH0_SECRET`, `SESSION_SECRET`) now use `generateState(32)` instead of a static placeholder
- Map iteration over `EnvValues` is now sorted for deterministic config file output
- Spring Boot YAML config uses `gopkg.in/yaml.v2` with `buildNestedMap()` to correctly produce nested YAML from dot-delimited keys (e.g. `okta.oauth2.issuer`)
- All detection-related string literals replaced with the `auth0.DetectionSub` constant
- Output switched from `fmt.Printf` to `cli.renderer.Infof` for consistency
- `m2m` removed from the application type selection list
- Fixed `"::none"` error when only `--api` was selected (config key lookup now correctly skipped for API-only flows)
- Trailing periods removed from section comment dividers in `internal/auth0/quickstart.go`

---

### References

- DXCDT-1554 — review comment fixes for `setup-experimental`
- DXCDT-1555 — tool detection module and CLI flow refinement

---

### Testing

**Manual — existing project:**
```bash
cd /path/to/react-vite-project
auth0 quickstarts setup-experimental
# Expected: detects react/spa/vite/5173, shows summary, confirms, skips re-prompting those fields
```

**Manual — empty directory:**
```bash
mkdir empty-test && cd empty-test
auth0 quickstarts setup-experimental
# Expected: no detection panel, prompts for type -> framework -> build tool -> port -> name
```

**Manual — API-only:**
```bash
auth0 quickstarts setup-experimental --api --identifier https://my-api
# Expected: prompts to select an existing app, creates ResourceServer, no app client created
```

**Manual — App + API:**
```bash
auth0 quickstarts setup-experimental --app --api --name "My App"
# Expected: creates client, creates ResourceServer named "My App-API"
```

Unit tests for `DetectProject` covering each of the 14 signal rules: N/A (to be added in a follow-up — mocking `os.ReadFile` / `os.Stat` for the full matrix is tracked separately).

---

### Checklist

- [x] I have added documentation for all new/changed functionality (command `--help` text and flag descriptions updated inline)


ramya18101:
@KartikJha , Please fix the [linters](https://github.com/auth0/auth0-cli/actions/runs/24133919436/job/70417029468?pr=1476) errors

docs/auth0_quickstarts_setup-experimental.md
```diff
@@ -0,0 +1,72 @@
...
+      --callback-url string     Override the allowed callback URL for the application
+      --framework string        Framework to configure (e.g., react, nextjs, vue, express)
+      --identifier string       Unique URL identifier for the API (audience), e.g. https://my-api
+      --logout-url string       Override the allowed logout URL for the application
+      --name string             Name of the Auth0 application
+      --offline-access          Allow offline access (enables refresh tokens)
+      --port int                Local port the application runs on (default varies by framework, e.g. 3000, 5173)
+      --scopes string           Comma-separated list of permission scopes for the API
+      --signing-alg string      Token signing algorithm: RS256, PS256, or HS256 (leave blank to be prompted interactively)
+      --token-lifetime string   Access token lifetime in seconds (default: 86400 = 24 hours) (default "86400")
```
ramya18101: Highlight that these flags are for the API

internal/auth0/quickstart.go
```diff
@@ -174,3 +173,516 @@ func (q Quickstarts) Stacks() []string {
...
+	"spa:react:vite": {
+		EnvValues: map[string]string{
+			"VITE_AUTH0_DOMAIN":    DetectionSub,
+			"VITE_AUTH0_CLIENT_ID": DetectionSub,
+		},
+		RequestParams: RequestParams{
+			AppType:           "spa",
+			Callbacks:         []string{"http://localhost:5173/callback"},
+			AllowedLogoutURLs: []string{"http://localhost:5173"},
+			WebOrigins:        []string{DetectionSub},
```
ramya18101: As discussed earlier, the port value should be prompted with default values like 5173 or 4200, as it varies depending on the configuration.

internal/cli/quickstart_detect.go
```diff
@@ -0,0 +1,584 @@
...
+	if hasDep(earlyDeps, "@ionic/angular") {
+		result.Framework = "ionic-angular"
+		result.Type = "native"
+		result.Detected = true
+		return result
+	}
+	if hasDep(earlyDeps, "@ionic/react") {
+		result.Framework = "ionic-react"
+		result.Type = "native"
+		result.BuildTool = "vite"
```
ramya18101: Instead of static build tool, try to fetch the build tool details from other files like .vite.config.ts

ramya18101:
- Colouring of newly created files/apps
- Fix the Prompt to confirm with name
- Port also should follow the name process
- Format the output msg
- Fix testcases for the main flow of each scenario which tests .env content and the requestparams Info
- Refactor the consts and help functions into 1 package instead of depending on auth package

ramya18101 (changes requested):

internal/display/display.go
```diff
@@ -65,6 +65,20 @@ func (r *Renderer) Infof(format string, a ...interface{}) {
 	fmt.Fprintf(r.MessageWriter, format+"\n", a...)
 }
 
+func (r *Renderer) InfofNoSpace(format string, a ...interface{}) {
```
ramya18101: `InfofNoSpace` is not used at anywhere

internal/cli/quickstart_detect.go
```diff
@@ -0,0 +1,1044 @@
...
+			// Flutter create (default) has included web/ since Flutter 2.10, so web/ alone
+			// is not a reliable signal for web-only intent.
+			// Native intent is signalled by android/ or ios/ platform directories.
+			// Web intent is signalled by a "web:" key nested under the "flutter:" section
+			// in pubspec.yaml. Absence of both signals defaults to native.
+			switch {
+			case dirExists(dir, "android") || dirExists(dir, "ios"):
+				result.Framework = "flutter"
+				result.Type = "native"
+				result.BundleID = readMobileBundleID(dir)
```
ramya18101, line 88: Can this case go along with the default case

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+				case detection.Detected:
+					noInputMode := !canPromptFlag
+					if len(detection.AmbiguousCandidates) > 1 {
+						// Multiple package.json deps matched - show partial summary and ask user to disambiguate.
+						cli.renderer.InfofBullet("Detected in current directory")
+						cli.renderer.InfofBullet("Framework: %s", "Could not be determined")
+						cli.renderer.InfofBullet("App type: %s", detectionFriendlyAppType(detection.Type))
+						cli.renderer.InfofBullet("App name: %s", detection.AppName)
+						if detection.Port > 0 {
+							cli.renderer.InfofBullet("Port: %d", detection.Port)
```
ramya18101: We don't need to display the app name and port as the detected values,
We can prompt the details by giving the default value

internal/cli/quickstart_detect.go
```diff
@@ -0,0 +1,1044 @@
...
+var portPattern = regexp.MustCompile(`"?port"?\s*:\s*(\d{4,5})`)
+
+// mobileTFMRegex matches .NET mobile Target Framework Monikers (net*-android or net*-ios).
+// A bare substring match on "-android" / "-ios" can produce false positives on package
+// names such as "Newtonsoft.Json-ios" or condition attributes. Requiring the leading
+// net<major>.<minor> prefix eliminates those false positives.
+var mobileTFMRegex = regexp.MustCompile(`net\d+\.\d+-(?:android|ios)`)
+
+// extractPortFromContent returns the first port number found in content, or 0 if none found.
+func extractPortFromContent(content string) int {
```
ramya18101: As discussed, we are not going to read the port values from the files, 
Instead use the pre-defined default values shared in the Command Redesign Spec doc - v. 6
Remove these additional code

internal/cli/quickstart_detect.go
```diff
@@ -0,0 +1,1044 @@
...
+		candidates = append(candidates, detectionCandidate{framework: "express", qsType: "regular", port: 3000})
+	}
+	if hasDep(deps, "hono") {
+		candidates = append(candidates, detectionCandidate{framework: "hono", qsType: "regular", port: 3000})
+	}
+	if hasDep(deps, "fastify") {
+		candidates = append(candidates, detectionCandidate{framework: "fastify", qsType: "regular", port: 3000})
+	}
+	return candidates
+}
```
ramya18101, line 308: recheck on the port values, as for all the non-native app types we are having the same default port

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+						if noInputMode || prompt.ConfirmWithDefault("Do you want to proceed with the detected values?", true) {
+							inputs = applyDetectionToInputs(inputs, detection)
+							if inputs.Framework == "" {
+								inputs.Framework = detection.Framework
+							}
+						}
+					}
+				default:
+					// No detection signal found - notify the user and pre-fill name from directory.
+					if !canPromptFlag && inputs.Type == "" {
```
ramya18101: Why are we checking the `inputs.Type` too..
As we are returning the error that highlights about `--no-input`

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+				return fmt.Errorf("failed to get quickstart configuration: %w", err)
+			}
+			inputs = updatedInputs
+			if inputs.App && wasAutoSelected {
+				cli.renderer.Infof("Auto-selected build tool %q for %s/%s (no exact match for 'none')", inputs.BuildTool, inputs.Type, inputs.Framework)
+			}
+
+			// -- Step 3b: Collect application name --.
+			if inputs.App {
+				if !cmd.Flags().Changed("name") {
```
ramya18101: Instead of manually checking  cmd.Flags().Changed("name") use `setupExpName.IsSet()` in all the places

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+				typeFromFlag := cmd.Flags().Changed("type")
+				frameworkFromFlag := cmd.Flags().Changed("framework")
+
+				switch {
+				case inputs.Type == "m2m":
+					// M2M apps have no framework or port; skip detection entirely so that
+					// signal files in the directory cannot override the explicit --type flag.
+					if inputs.Name == "" {
+						inputs.Name = detection.AppName
+					}
```
ramya18101: Can we do this check before doing the DetectProject as it is not needed for m2m apps..

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+				case typeFromFlag && frameworkFromFlag:
```
ramya18101: Why do we check for only these 2 flags..?

internal/cli/quickstart_detect.go
```diff
@@ -0,0 +1,1044 @@
...
+// multiple package.json deps match and the framework cannot be determined uniquely.
+type DetectionResult struct {
+	Framework           string
+	Type                string   // "spa" | "regular" | "native".
+	BuildTool           string   // "vite" | "maven" | "gradle" | "composer" | "" (NA).
+	Port                int      // 0 means no applicable default.
+	AppName             string   // Basename of the working directory.
+	BundleID            string   // Package/bundle ID for native apps (e.g. "com.example.myapp"); empty if not found.
+	Detected            bool     // True if any signal file matched.
+	AmbiguousCandidates []string // Set when >1 package.json dep matched.
```
ramya18101: Rename as AmbiguousFrameworks

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+	return frameworks
+}
+
+// getQuickstartConfigKey resolves remaining missing prompts for App and API creation
+// and returns the config map key for the selected framework.
+// App/API selection and project detection are handled by the caller before this is invoked.
+func getQuickstartConfigKey(inputs SetupInputs) (string, SetupInputs, bool, error) {
+	// Handle application creation inputs.
+	if inputs.App {
+		// Validate --type if provided (Bug 12).
```
ramya18101: Explain this?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+
+		// Prompt for --type if not provided.
+		if inputs.Type == "" {
+			q := prompt.SelectInput("type", "Select the application type", "", validTypes, "spa", true)
+			if err := prompt.AskOne(q, &inputs.Type); err != nil {
+				return "", inputs, false, fmt.Errorf("failed to select application type: %w", err)
+			}
+		}
+
+		// M2M apps have no framework, port, or callback URLs (Bug 6).
```
ramya18101: Also Please explain this..?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+				return "", inputs, false, fmt.Errorf("failed to select framework: %w", err)
+			}
+		}
+
+		// Resolve port from framework default before prompting (Bug 11).
+		// The spec says "--port: default value used if not given", so we never prompt.
+		if inputs.Port == 0 {
+			inputs.Port = defaultPortForFramework(inputs.Framework)
+			// Port stays 0 for native apps (react-native, expo, flutter) - no port needed.
+		}
```
ramya18101: Isn't this handled again under `- Step 3d: Prompt for port if not explicitly set --.`

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+	}
+
+	// Config key is only meaningful when an app is being created.
+	if !inputs.App {
+		return "", inputs, false, nil
+	}
```
ramya18101: Do we really need this check..?
As app check is already happened in the start!

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+			if len(frameworks) == 0 {
+				return "", inputs, false, fmt.Errorf("no frameworks available for type %q", inputs.Type)
+			}
+			q := prompt.SelectInput("framework", "Select the framework", "", frameworks, frameworks[0], true)
+			if err := prompt.AskOne(q, &inputs.Framework); err != nil {
+				return "", inputs, false, fmt.Errorf("failed to select framework: %w", err)
+			}
+		}
+
+		// Resolve port from framework default before prompting (Bug 11).
```
ramya18101: Why are we highlighting the bugs..?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+					return pi < pj
+				}
+				return candidates[i] < candidates[j]
+			})
+			configKey = candidates[0]
+			// Update inputs.BuildTool so the caller can notify the user of the auto-selection.
+			parts := strings.SplitN(configKey, ":", 3)
+			if len(parts) == 3 {
+				inputs.BuildTool = parts[2]
+			}
```
ramya18101: Instead of this complicated logic, Can we enhance the `frameworksForType` logic to store the supported buildTools for each Framework!

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+		}
+	}
+	sort.Strings(frameworks)
+	return frameworks
+}
+
+// getQuickstartConfigKey resolves remaining missing prompts for App and API creation
+// and returns the config map key for the selected framework.
+// App/API selection and project detection are handled by the caller before this is invoked.
+func getQuickstartConfigKey(inputs SetupInputs) (string, SetupInputs, bool, error) {
```
ramya18101: Divide it into functions as fetchNecessaryInputs and GetQsConfigKey

and since both `inputs` was used in both input and output params of that func, Please wisely use reference(*)

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+			if inputs.API && !inputs.App {
+				// Collect API name if not already set (pre-fill from CWD folder name).
+				if inputs.Name == "" && !cmd.Flags().Changed("name") {
+					cwd, _ := os.Getwd()
+					defaultName := filepath.Base(cwd)
+					if defaultName == "" || defaultName == "." {
+						defaultName = "my-api"
+					}
+					if canPromptFlag {
+						q := prompt.TextInput("name", "Application Name", "Name for the Auth0 API", defaultName, true)
```
ramya18101: Why didn't we use existing flags here..?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+					return err
+				}
+
+				// If the flag was not set, prompt interactively; fall back to 86400 in non-interactive mode.
+				if inputs.TokenLifetime == "" {
+					if canPromptFlag {
+						defaultLifetime := "86400"
+						q := prompt.TextInput("token-lifetime", "Access token lifetime (seconds)",
+							"How long access tokens remain valid (default: 86400 = 24 hours)", defaultLifetime, true)
+						if err := prompt.AskOne(q, &inputs.TokenLifetime); err != nil {
```
ramya18101: Instead of using `prompt.AskOne` and `cmd.Flags().Changed` use our in-bulit flag methods

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+	setupExpBuildTool.RegisterString(cmd, &inputs.BuildTool, "none")
+	setupExpPort.RegisterInt(cmd, &inputs.Port, 0)
+	setupExpCallbackURL.RegisterString(cmd, &inputs.CallbackURL, "")
+	setupExpLogoutURL.RegisterString(cmd, &inputs.LogoutURL, "")
+	setupExpWebOriginURL.RegisterString(cmd, &inputs.WebOriginURL, "")
+
+	// API flags.
+	setupExpAPI.RegisterBool(cmd, &inputs.API, false)
+	setupExpIdentifier.RegisterString(cmd, &inputs.Identifier, "")
+	setupExpAudience.RegisterString(cmd, &inputs.Audience, "")
```
ramya18101: Why do we have `identifier` and `audience` as two seperate flags..
Aren't these both same..?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+
+			if inputs.API {
+				// Prompt for the identifier if not explicitly provided via flag.
+				if !cmd.Flags().Changed("identifier") && !cmd.Flags().Changed("audience") {
+					// Compute a suggested default without pre-populating inputs.Identifier.
+					defaultID := inputs.Identifier
+					if defaultID == "" {
+						defaultID = inputs.Audience
+					}
+					if defaultID == "" && inputs.Name != "" {
```
ramya18101: WHy are we having 2 flags..?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+		return 3000
+	}
+}
+
+// validateAPIIdentifier returns an error if identifier is not a valid http:// or https:// URL.
+func validateAPIIdentifier(identifier string) error {
+	// Err != nil from url.Parse only fires on malformed percent-encoding; the
+	// host check catches bare schemes like "http://" that Parse accepts without error.
+	// u.User != nil rejects URLs with embedded credentials (e.g. http://user:pass@host).
+	u, err := url.Parse(identifier)
```
ramya18101: Use the similar existing logic to handle validation: https://github.com/auth0/auth0-cli/blob/18e479ef146a71a3652717ba6b6a248ffe1fcd75/internal/cli/apis.go#L606

internal/auth0/quickstart.go
```diff
@@ -174,3 +185,637 @@ func (q Quickstarts) Stacks() []string {
 
 	return stacks
 }
+
+const (
+	// DetectionSub is replaced at runtime with baseURL+CallbackPath ("/callback" by default).
+	DetectionSub = "DETECTION_SUB"
+	// DetectionSubBase is replaced at runtime with just the baseURL (no path suffix).
+	// Use this for SPA callback/logout URLs where the path is the app root.
+	DetectionSubBase = "DETECTION_SUB_BASE"
```
ramya18101: Rename as DetectionSubAsBase

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+		switch cb {
+		case auth0.DetectionSub:
+			callbacks[i] = baseURL + callbackPath
+		case auth0.DetectionSubBase:
+			callbacks[i] = baseURL
+		}
+	}
+	for i, u := range logoutURLs {
+		if u == auth0.DetectionSub || u == auth0.DetectionSubBase {
+			logoutURLs[i] = baseURL
```
ramya18101: WHy are we checking  `u == auth0.DetectionSubBase` as we never set in the whole codebase as `DetectionSubBase`, we only used `DetectionSub`

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+	updatedEnvValues := make(map[string]string)
+
+	for key, value := range envValues {
+		if value != auth0.DetectionSub && value != auth0.DetectionSubBase {
+			updatedEnvValues[key] = value
+			continue
+		}
+
+		switch key {
+		case "VITE_AUTH0_DOMAIN", "AUTH0_DOMAIN", "domain", "NUXT_AUTH0_DOMAIN",
```
ramya18101: Divide this function into: `replaceDetectionSub` and `resolveDetectionSubValue`
as it currently does 2 operations

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+	envValues := config.EnvValues
+	if inputs.API && inputs.Identifier != "" && config.AudienceVar != "" {
+		envValues = make(map[string]string, len(config.EnvValues)+1)
+		for k, v := range config.EnvValues {
+			envValues[k] = v
+		}
+		envValues[config.AudienceVar] = inputs.Identifier
+	}
+
+	envFileName, _, err := GenerateAndWriteQuickstartConfig(&config.Strategy, envValues, cli.tenant, client, inputs.Port)
```
ramya18101: If we are not using the 2nd field from the output param of `GenerateAndWriteQuickstartConfig` then why are we even sending it..?

internal/cli/quickstarts.go
```diff
@@ -656,3 +770,1211 @@ func setupQuickstartCmd(cli *cli) *cobra.Command {
...
+			s = strings.TrimSpace(s)
+			if s != "" {
+				v := s
+				apiScopes = append(apiScopes, management.ResourceServerScope{Value: &v})
+			}
+		}
+		if len(apiScopes) > 0 {
+			rs.Scopes = &apiScopes
+		}
+	}
```
ramya18101, line 1219: Take the similar reference from : https://github.com/auth0/auth0-cli/blob/18e479ef146a71a3652717ba6b6a248ffe1fcd75/internal/cli/apis.go#L693


and checkout the whole createAPI logic from the `apis.go/createAPICmd`

ramya18101 (changes requested):

internal/cli/quickstart_detect.go
KartikJha:
```

Current code: native dirs take priority over web target.                                                                                                                                      
  Collapsed version: web target takes priority over native dirs.
                                                                                                                                                                                                
  // Collapsed — web target wins even if android/ or ios/ exist
  switch {
  case pubspecHasFlutterWebTarget(data):
      result.Framework = "flutter-web"
      result.Type = "spa"
  default:
      result.Framework = "flutter"
      result.Type = "native"
      result.BundleID = readMobileBundleID(dir)
  }

  A Flutter project can legitimately have both — flutter create --platforms=android,ios,web produces a project with native dirs and a web: section in pubspec. The current code intentionally
  treats native dirs as the stronger signal, since a developer running setup-experimental on such a project is more likely targeting mobile Auth0 integration.

  So collapsing is valid only if you're okay with flipping that priority. If the intent is "native dirs beat web target when both exist", the explicit first case is load-bearing and should
  stay.
  
  ```
  
  
  
  @ramya18101 most probably in this scenario we want to force the native scenario.  Let me know if we still want to target web first then I can collapse the switch.
  
  
  
  cc @arpit-jn @duedares-rvj

internal/cli/quickstart_detect.go
KartikJha: @ramya18101 as discussed in the call time complexity of using this function for extracting the port from signal files is not greater than usability of auto port detection as it runs on client machine and will make the command more AI tooling friendly.

The fallback behavior already covers the concern here. detectPortFromConfig returns defaultPort whenever extraction   
  fails — so using predefined defaults is what the code does when no port is found in the config file. The file read is 
  just a best-effort check before that.                                                                                 
                  
The reason to keep it: inputs.Port feeds directly into the Auth0 callback URL registration (baseURL :=
  fmt.Sprintf("http://localhost:%d", inputs.Port) at line 635). If a developer has customized their vite.config.ts to
  port: 4000 and we register http://localhost:5173, Auth0 will reject the redirect at login. Detecting the actual port
  prevents a silent misconfiguration that's painful to debug.

Removing extractPortFromContent would mean the fallback default is always used, trading a small file read for a
  potential broken Auth0 callback URL. Given the function already has the safety net of falling back to the default on
  any failure, I'd keep it.

internal/cli/quickstart_detect.go
KartikJha: As discussed this comment is not an issue main issue is [here](https://github.com/auth0/auth0-cli/pull/1476#discussion_r3195167484)

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: @ramya18101 

The inputs.Type == "" check determines which error the user sees when detection fails in --no-input
  mode. There are two sub-cases that reach default:                                                     
   
  1. Neither --type nor --framework passed — inputs.Type is empty, so we error here with the full       
  guidance message.                                         
  2. --type passed but not --framework — inputs.Type is non-empty, so we skip this error and fall
  through to line 928, which emits the more specific --framework is required when --type is spa message.

  Note that a user passing both --type and --framework never reaches default — they match the
  typeFromFlag && frameworkFromFlag case earlier and bypass detection entirely.

  So the two guards together cover all missing-flag combinations: this one handles the fully-unknown
  case, line 928 handles the partially-specified case. Dropping inputs.Type == "" would collapse both
  into the generic error, which is less helpful when the user only forgot --framework.
  
  cc @arpit-jn @duedares-rvj @bkiran6398

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: These flags are checked because no other detected values apart from `bundleId` should be applied to inputs, also it should come after `DetectProject` because bundleId needs to be read from the signal files, also since these framework and types were not detected rather read from user input flags we need a separate case to handle it.

internal/cli/quickstart_detect.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: The two checks serve different execution paths and are not redundant.                                           
                                                                                                                    
  The check at line 1482 (inside getQuickstartConfigKey) is the non-interactive fallback — when --no-input is set,
  the Step 3d block is skipped entirely (guarded by canPromptFlag), so getQuickstartConfigKey is what ensures the   
  port gets a framework default before the config key is resolved.

  The check at line 972 (inside Step 3d) pre-fills the interactive prompt default — without it, the prompt would
  display 0 for the user to edit instead of the framework-appropriate value (e.g. 5173 for Vite). It's not setting
  the final value; it's setting the starting value shown to the user.

  So removing either one would be a regression: removing line 1482 breaks --no-input mode, removing line 972 breaks
  the interactive prompt UX.

ramya18101: So, re-check and remove the redundant logic as it already handled here

internal/cli/quickstarts.go
KartikJha: Good catch. Implemented the early return pattern — the redundant if inputs.App wrapper and the unreachable if !inputs.App check have been removed. The function
  now starts with:                                                                                                                                                 
   
  if !inputs.App {                                                                                                                                                 
      return "", inputs, false, nil                         
  }

  and the rest of the logic sits flat without an extra nesting level.

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstart_detect.go
KartikJha: Updated

internal/cli/quickstart_detect.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Done — introduced auth0.FrameworkBuildTools, a package-level map built from QuickstartConfigs at startup, keyed by
   "type:framework" with the supported build tools sorted alphabetically. The dynamic search-and-sort block in      
  getQuickstartConfigKey is replaced with a single map lookup before configKey is built:                          
                                                                                                                    
  if tools := auth0.FrameworkBuildTools[inputs.Type+":"+inputs.Framework]; len(tools) > 0 {
      buildToolKey = tools[0]
      inputs.BuildTool = buildToolKey
      wasAutoSelected = true
  }

  Kept it in auth0/quickstart.go alongside QuickstartConfigs rather than inside frameworksForType since it's derived
   directly from the config map and serves a different concern — build tool resolution rather than prompting.

internal/cli/quickstarts.go
KartikJha: Done — extracted resolveSetupInputs (which handles type validation/prompting, framework prompting, port default,  
  and build tool resolution) out of getQuickstartConfigKey, keeping the latter focused on flow control and key      
  construction.
                                                                                                                    
  On the pointer suggestion: SetupInputs is a plain struct with only value-type fields (string, bool, int) plus one
  map. In Go, the idiomatic pattern for "transform and return a struct" is pass-by-value with a named return — it
  avoids aliasing, keeps callers from observing partial mutations on error paths, and makes the data flow explicit
  in the signature. Since the struct is passed by value, any modifications inside resolveSetupInputs operate on a
  local copy and do not affect the caller's original struct — the updated inputs are explicitly returned and the
  caller decides what to do with them. A pointer would be appropriate if the struct were large enough to make
  copying expensive, or if we needed shared mutable state across goroutines — neither applies here. The current
  (SetupInputs, bool, error) return signature is the right call.

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/auth0/quickstart.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: The second output is the path of the envFile and is used in Test Suites to check for temp env file existence and overall makes the test suite more reliable.

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
KartikJha: --audience is an intentional alias for --identifier to accommodate developers who are more familiar with Auth0's "audience"
  terminology (used widely in Auth0 docs and SDK config). Both resolve to the same value — the resource server identifier. The else
   if inputs.Identifier == "" { inputs.Identifier = inputs.Audience } line is exactly what handles that aliasing. The flag help
  text already documents this: "Alias for --identifier (unique audience URL for the API)".
  
  cc @arpit-jn @duedares-rvj

internal/cli/quickstarts.go
KartikJha: Assuming this is related to audience and identifier, it has been addressed [here](https://github.com/auth0/auth0-cli/pull/1476#discussion_r3205629740)

internal/cli/quickstarts.go
KartikJha: Updated

internal/cli/quickstarts.go
ramya18101: We can make use of  Alias property of Flags, Take reference from: https://github.com/auth0/auth0-cli/blob/main/internal/cli/attack_protection.go#L11

internal/cli/quickstarts.go
ramya18101: Using underscores in the main files solely to satisfy test suites doesn’t seem ideal

kushalshit27 (approved):

ramya18101 (approved):

PR #1522, opened 2026-05-18
feat: Support third party app security params

bkiran6398:
### 🔧 Changes

Adds CLI support for Auth0's Third-Party Client security improvements.

**New flags on `apps create`:**
- `--is-first-party` (`-f`): Whether the application is a first-party client (true) or third-party client (false). Default: `true`
- `--third-party-security-mode` (`-s`): Set security mode to `strict` or `permissive`
- `--redirection-policy` (`-y`): Set to `allow_always` or `open_redirect_protection`

**New flags on `apps update`:**
- `--is-first-party` (`-f`): Update whether the application is first-party or third-party
- `--third-party-security-mode` (`-s`): Update the security mode
- `--redirection-policy` (`-y`): Update the redirection policy

**Display updates:**
- `apps show` and `apps list` now display `IS FIRST PARTY`, `THIRD PARTY SECURITY MODE` and `REDIRECTION POLICY` fields when set

**Terraform fetcher fix:**
- Handles `default_for` client grants correctly during `auth0 tf generate`. Grants with `default_for` (which lack a `client_id`) now produce a meaningful resource name (`default_for_third_party_clients_<audience>`) instead of a broken `_<audience>` name.

**Behavioral notes:**
- Default grant types are not applied for third-party apps (the API manages defaults)
- Logout URL prompts are suppressed for third-party apps (not applicable)
- Update uses `IsSet()` guards so these fields are only sent when explicitly provided

### 📚 References

- API Docs: https://auth0.com/docs/api/management/v2/clients/post-clients

### 🔬 Testing

- Unit test added: `TestClientGrantResourceFetcher_FetchData` — verifies `default_for` grants produce correct resource names
- Integration test added: test case 051 creates a regular app with `--third-party-security-mode strict --redirection-policy open_redirect_protection` and validates JSON output

#### Manual testing:
 - Verified CRUD operation on 3rd party apps using new flag values
 - Verified terraform generation for client grants with default_for value
 
### 📝 Checklist

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -186,13 +186,13 @@ var (
 		Name:      "Third Party Security Mode",
 		LongForm:  "third-party-security-mode",
 		ShortForm: "s",
-		Help:      "Security mode for third-party clients: 'strict' or 'permissive'.",
+		Help:      "Security mode for third-party clients: 'strict' or 'permissive'. Require --is-first-party=false",
 	}
 	appRedirectionPolicy = Flag{
 		Name:      "Redirection Policy",
 		LongForm:  "redirection-policy",
 		ShortForm: "y",
-		Help:      "Controls whether Auth0 redirects users to the application's callback URL on authentication errors or in email verification flows: 'allow_always' or 'open_redirect_protection'.",
+		Help:      "Controls whether Auth0 redirects users to the application's callback URL on authentication errors or in email verification flows: 'allow_always' or 'open_redirect_protection'. Require --is-first-party=false",
 	}
 )
 
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -176,6 +176,12 @@ var (
 		ShortForm: "p",
 		Help:      "Comma-separated list of enabled token exchange types for this client. Possible values: custom_authentication, on_behalf_of_token_exchange.",
 	}
+	appIsFirstParty = Flag{
+		Name:      "Is First Party",
+		LongForm:  "is-first-party",
+		ShortForm: "f",
+		Help:      "Whether the application is a first-party client (true) or third-party client (false).",
+	}
 	appThirdPartySecurityMode = Flag{
 		Name:      "Third Party Security Mode",
 		LongForm:  "third-party-security-mode",
@@ -446,6 +452,7 @@ func createAppCmd(cli *cli) *cobra.Command {
 		RefreshToken             string
 		ResourceServerIdentifier string
 		AllowAnyProfileOfType    []string
+		IsFirstParty             bool
 		ThirdPartySecurityMode   string
 		RedirectionPolicy        string
 	}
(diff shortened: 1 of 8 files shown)


internal/cli/apps.go
```diff
@@ -593,11 +609,21 @@ func createAppCmd(cli *cli) *cobra.Command {
...
 			}
 
+			if inputs.ThirdPartySecurityMode != "" {
+				a.ThirdPartySecurityMode = &inputs.ThirdPartySecurityMode
+			}
+			if inputs.RedirectionPolicy != "" {
+				a.RedirectionPolicy = &inputs.RedirectionPolicy
+			}
+			if isThirdPartyApp {
+				a.IsFirstParty = auth0.Bool(false)
```
ramya18101: Do we manually need to set it is as false, while initializing `a.ThirdPartySecurityMode` and what happens if the users want to update the app to an non-3rd party app .. then are we going to set back `a.IsFirstParty = auth0.Bool(true)` ..?

internal/cli/terraform_fetcher.go
```diff
@@ -274,8 +274,14 @@ func (f *clientGrantResourceFetcher) FetchData(ctx context.Context) (importDataL
 		}
 
 		for _, grant := range grants.ClientGrants {
+			var resourceName string
+			if grant.GetDefaultFor() != "" {
+				resourceName = "auth0_client_grant." + sanitizeResourceName("default_for_"+grant.GetDefaultFor()+"_"+grant.GetAudience())
```
ramya18101: Why are we handling this case separately and adding a prefix ..?

test/integration/apps-test-cases.yaml
```diff
@@ -376,3 +376,13 @@ tests:
   050 - given a resource server app, it successfully deletes the app:
     command: auth0 apps delete $(./test/integration/scripts/get-resource-server-app-id.sh) --force
     exit-code: 0
+
+  051 - it successfully creates a regular app with third-party-security-mode and redirection-policy and outputs in json:
```
ramya18101: Also please add a case with update..!

ramya18101 (changes requested):

test/integration/apps-test-cases.yaml
bkiran6398: Added now.

internal/cli/apps.go
bkiran6398: Added a new flag to allow users explicitly set the `is-first-party` value.

internal/cli/terraform_fetcher.go
bkiran6398: Grants with `default_for` have no `client_id`; hence, `default_for` is used as the prefix instead of `client_id`.

ramya18101:
Please add a friendly helping warning log, if any users try to just set only `third-party` related fields without manually setting `is-first-party` to false..

Let the users find a good help log on those scenario's as they are inter-dependent

ramya18101 (approved):

PR #1525, opened 2026-05-19
Install claude skills with auth0 cli

KartikJha:
### 🔧 Changes                                                                                                                
                                                                                                                                
  Introduces a new `internal/ai/skills` package that provides an agent registry and detection layer for AI coding agents. This
  is the foundation for the `auth0 ai skills install` command, enabling the CLI to discover which agents are present on a user's
   machine and resolve the correct skills directory for each.
                                                                                                                                
  **New types and functions:**

  - `AgentConfig` — describes a single AI coding agent with its ID, display name, global/project skills directories, and
  detection hints (filesystem markers and binary names)
  - `AgentConfig.IsInstalled()` — returns `true` if any marker path exists on disk or any binary is found in `$PATH`
  - `SupportedAgents` — registry of 20 agents: Claude Code, Cursor, GitHub Copilot, Gemini CLI, Roo Code, Goose, OpenCode,
  Codex, Windsurf, Continue, Amp, Junie, Kiro CLI, Cline, Augment, AiderDesk, Warp, OpenHands, Trae, and a Universal fallback
  - `DetectedAgents()` — returns the subset of `SupportedAgents` that are installed, plus the Universal agent; result is cached
  via `sync.Once`
  - `FastPriorityAgents()` — returns detected agents sorted by priority order (Claude Code → Cursor → GitHub Copilot → Gemini
  CLI → others → Universal last), used by the `--fast` install mode

  **Cleanup:**

  - Removed three dead flag variables (`qsType`, `qsAppName`, `qsPort`) from `quickstarts.go` that were no longer referenced
  - Minor comment capitalisation fixes in `quickstart_detect.go` and `quickstart.go`

  ### 🔬 Testing

  All new behaviour is covered by unit tests in `internal/ai/skills/agent_test.go`:

  - `TestIsInstalled` — marker path existence, binary lookup, empty-string skipping, fallback from marker miss to binary hit
  - `TestSupportedAgents` — non-empty registry, unique IDs, non-empty skill dirs, presence of universal, agents with nil
  detection markers
  - `TestDetectedAgents` — universal always present, deterministic caching, all results are subsets of `SupportedAgents`
  - `TestFastPriorityAgents` — universal last, no duplicates, all detected agents present, priority ordering invariant, length
  equality

  Run with:
  go test ./internal/ai/skills/...

  No external services or credentials are required; `IsInstalled` checks are exercised using `t.TempDir()` and `t.Setenv("PATH",
   ...)`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,4 +47,3 @@ out/
 # LLM files
 .remember/
 
-.gopath
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ out/
 
 # LLM files
 .remember/
+
+.gopath
--- a/internal/ai/skills/agent.go
+++ b/internal/ai/skills/agent.go
@@ -217,8 +217,8 @@ func init() {
 }
 
 var (
-	detectedAgentsOnce   sync.Once
-	detectedAgentsCache  []AgentConfig
+	detectedAgentsOnce  sync.Once
+	detectedAgentsCache []AgentConfig
 )
 
 // DetectedAgents returns the subset of SupportedAgents that are installed on this machine.
--- a/internal/ai/skills/agent.go
+++ b/internal/ai/skills/agent.go
@@ -7,18 +7,26 @@ import (
 	"sync"
 )
 
-// AgentConfig describes a single AI coding agent and where it reads skills from.
 type AgentConfig struct {
-	ID               string
-	DisplayName      string
-	GlobalSkillsDir  string
-	ProjectSkillsDir string
-	DetectMarkers    []string // Paths whose existence indicates the agent is installed (any one match is sufficient).
-	DetectBinaries   []string // Binary names to look up in PATH (any one match is sufficient).
+	ID                   string
+	DisplayName          string
+	GlobalSkillsDir       string
+	GlobalSkillsDirEnvVar string
+	ProjectSkillsDir      string
+	DetectMarkers         []string
+	DetectMarkerEnvVars   []string
+	DetectBinaries        []string
 }
 
-// IsInstalled reports whether the agent appears to be installed on this machine.
-// It returns true if any marker path exists or any binary is found in PATH.
+func (a AgentConfig) ResolvedGlobalSkillsDir() string {
+	if a.GlobalSkillsDirEnvVar != "" {
+		if v := os.Getenv(a.GlobalSkillsDirEnvVar); v != "" {
+			return filepath.Join(v, "skills")
+		}
+	}
+	return a.GlobalSkillsDir
+}
+
 func (a AgentConfig) IsInstalled() bool {
 	for _, marker := range a.DetectMarkers {
 		if marker == "" {
@@ -26,6 +34,16 @@ func (a AgentConfig) IsInstalled() bool {
 		}
 		if _, err := os.Stat(marker); err == nil {
 			return true
+		}
+	}
+	for _, envVar := range a.DetectMarkerEnvVars {
+		if envVar == "" {
+			continue
+		}
+		if v := os.Getenv(envVar); v != "" {
+			if _, err := os.Stat(v); err == nil {
+				return true
+			}
 		}
 	}
 	for _, binary := range a.DetectBinaries {
@@ -39,12 +57,14 @@ func (a AgentConfig) IsInstalled() bool {
 	return false
 }
 
-// SupportedAgents is the full list of AI coding agents that support the agentskills spec.
 var SupportedAgents []AgentConfig
 
 func init() {
 	home, _ := os.UserHomeDir()
 	if home == "" {
+		SupportedAgents = []AgentConfig{
+			{ID: "universal", DisplayName: "Universal", ProjectSkillsDir: filepath.Join(".agents", "skills")},
+		}
 		return
 	}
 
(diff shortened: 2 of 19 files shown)


internal/ai/skills/agent.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+}
+
+var (
+	detectedAgentsMu    sync.RWMutex
+	detectedAgentsDone  bool
+	detectedAgentsCache []AgentConfig
+)
+
+func DetectedAgents() []AgentConfig {
```
ramya18101, line 257: Do we need the mutex and resettable shared cache here at all? Since this runs in a CLI, agent detection will usually happen in a fresh process, so the main benefit seems to be re-detection within the same process,

internal/ai/skills/download.go
```diff
@@ -0,0 +1,410 @@
+package skills
```
ramya18101, line 1: Can we simplify this download flow substantially and make it closer to downloadAndUnzipSampleRepo? Since agent-skills is a public repo and this is a one-time install, I’m not sure we need all of: fetchCommitSHA, downloadViaGit, multiple archive fallbacks, and the atomic rename/cross-filesystem replacement logic. It feels like a single HTTP-based download from main plus extract/copy into place may be enough here.

internal/ai/skills/lock.go
```diff
@@ -0,0 +1,97 @@
+package skills
```
ramya18101, line 1: What concrete problem are we solving with cross-process locking here? auth0 ai skills install looks like a user-invoked, one-time install command, so concurrent installs seem like an edge case rather than an expected workflow. This adds platform-specific locking complexity that may not be justified unless we need to coordinate multiple processes updating the same install state.

ramya18101 (changes requested):

internal/ai/skills/skill_meta.go
```diff
@@ -0,0 +1,73 @@
...
+
+// SkillMeta holds the name and description extracted from a SKILL.md frontmatter.
+type SkillMeta struct {
+	Name        string `yaml:"name"`
+	Description string `yaml:"description"`
+}
+
+// ParseSkillMeta reads SKILL.md from skillDir and extracts the YAML frontmatter.
+// Returns an empty SkillMeta (no error) when the file has no valid frontmatter delimiters.
+func ParseSkillMeta(skillDir string) (SkillMeta, error) {
```
ramya18101, line 21: Should ParseSkillMeta() only treat frontmatter as valid when it appears at the start of SKILL.md? Right now any two --- lines anywhere in the file would be parsed as metadata.

internal/ai/skills/symlink.go
```diff
@@ -0,0 +1,180 @@
+package skills
```
ramya18101, line 1: Do we need all of this installation complexity here? For auth0 ai skills install, this seems to handle a lot of edge cases symlink vs copy mode, Windows junction fallback, temp-dir promotion, cross-filesystem fallback, and multiple validation state and I’m not sure all of those are required for the expected use case.

.gitignore
```diff
@@ -45,4 +45,5 @@ out/
 .github/copilot-instructions.md
 
 # LLM files
-.remember/
```
ramya18101, line 48: Un-necessary change

ramya18101 (changes requested):

internal/ai/skills/download.go
KartikJha: 1. fetchCommitSHA is needed for versioning of skills
2. The download needs to support the 3 types as git might not be present
3. Since we are covering updates as well this is a not a one might install
4. Atomic renames reduce the write complexity of [clone, sparse-checkout, place], [download, extract, place] * 2 
5. As discussed we are supporting multiple download types for resiliency, also extract/clone and copy need atleast 2 directories, present code implements that will error handling.

internal/ai/skills/agent.go
KartikJha: Yes, this is added to support re-detection for the same process. Making the download flow optimized

internal/ai/skills/lock.go
KartikJha: This encapsulates the skills-lock data structure, this is related to separation of concerns.

internal/ai/skills/symlink.go
KartikJha: Yes the symlinking allows for single copy and upgrade propagation, also we need to handle cross-filesysystem cases and windows cases as we supporting windows download as well.

internal/ai/skills/skill_meta.go
KartikJha: Since we control all SKILL.md files and enforce frontmatter at the start by convention, this isn't a practical concern. All skill files in this repo follow the ---/YAML/--- at the top structure, so parts[1] will always resolve to the correct block. We could add a strings.TrimSpace(parts[0]) != "" guard for strictness, but it's unnecessary overhead given the invariant we maintain.

ramya18101:
Take reference from : https://github.com/auth0/auth0-cli/blob/main/internal/cli/acul_app_scaffolding.go#L170

KartikJha:
> Take reference from : https://github.com/auth0/auth0-cli/blob/main/internal/cli/acul_app_scaffolding.go#L170


Hey @ramya18101 ,

Have incorporated the changes in https://github.com/auth0/auth0-cli/pull/1555

ramya18101:
Closing this PR but keeping the branch around, since we've cut a follow-up [PR](https://github.com/auth0/auth0-cli/pull/1555) that takes a simpler approach.

PR #1527, opened 2026-05-19
fix(keyring): clear stale access-token chunks instead of deleting the client secret

### LINKED ISSUE #1526: StoreAccessToken deletes the client secret instead of clearing stale access-token chunks
### Describe the problem

`internal/keyring/keyring.go`'s `StoreAccessToken` has a copy-paste error in the loop that is meant to clear previously-stored access-token chunks. Instead of deleting the indexed `secretAccessToken` chunks, it deletes the `secretClientSecret` key on every iteration:

```go
// First, clear any existing chunks to prevent concatenation issues.
for i := 0; i < secretAccessTokenMaxChunks; i++ {
    if err := keyring.Delete(secretClientSecret, tenant); err != nil { // <-- wrong key
        ...
    }
}
```

This was introduced in #1358.

### Expected behavior

The loop should delete the indexed access-token chunks (`fmt.Sprintf("%s %d", secretAccessToken, i)`), exactly as `DeleteSecretsForTenant` already does, and must not touch the client secret.

### Impact — two distinct failures

**1. Stale chunks → "authentication token is corrupted".**
Because old access-token chunks are never cleared, re-storing a token that spans fewer chunks than the previous one leaves trailing chunks behind. `GetAccessToken` concatenates them into a malformed JWT, `CheckAuthenticationStatus` fails `jwt.ParseInsecure`, and commands error with `authentication token is corrupted, please run: auth0 logout && auth0 login`.

**2. Machine (client-credentials) logins can never refresh.**
`RunLoginAsMachineSecret` stores the client secret and then calls `StoreAccessToken`, which immediately deletes that secret. Once the access token expires, `RegenerateAccessToken → GetClientSecret` fails with `failed to retrieve client secret from keyring`, so client-credentials sessions break at the first token expiry (~24h for the Management API) and require a brand-new `auth0 login` every time. Auto-renewal never works.

### Reproduction (machine auth)

```
auth0 login --domain <tenant> --client-id <id> --client-secret <secret>
# works for ~24h, then on next expiry:
auth0 apps list
# -> failed to fetch access token using client credentials ... please re-authenticate
```

### Environment

- auth0-cli **v1.30.0** (also confirmed present on `main` and at the `v1.30.0` tag)
- macOS Keychain via go-keyring, but the bug is OS-independent

### Fix

One-line correction + regression tests (deletes the indexed access-token chunks; leaves the client secret intact). PR incoming.

sparkling:
## Description

`StoreAccessToken`'s "clear existing chunks" loop deleted the `secretClientSecret` key on every iteration instead of the indexed `secretAccessToken` chunks (introduced in #1358). This caused two failures:

1. **Stale chunks → corrupted token.** Old access-token chunks were never cleared, so re-storing a shorter token left trailing chunks that concatenate into a malformed JWT → `authentication token is corrupted`.
2. **Machine logins couldn't refresh.** The client secret was deleted immediately after a machine login stored it, so `RegenerateAccessToken → GetClientSecret` failed and client-credentials sessions broke at the first token expiry (and never auto-renewed).

The loop now deletes the indexed access-token chunks (matching `DeleteSecretsForTenant` and the store/get loops) and leaves the client secret untouched.

Fixes #1526

## Testing

Added two regression tests in `internal/keyring/keyring_test.go`:
- overwriting a multi-chunk token with a single-chunk token returns exactly the new token (no stale chunks left over)
- `StoreAccessToken` does not delete the client secret

Both **fail on `main`** and pass with this change. Also verified end-to-end against a live tenant: a machine login followed by a forced token expiry now auto-renews silently instead of erroring with `please re-authenticate`.

## Checklist

- [x] Regression tests added
- [x] `go test ./internal/keyring/...` passes
- [x] Change is minimal and matches existing patterns in the file

### DIFF
--- a/internal/keyring/keyring.go
+++ b/internal/keyring/keyring.go
@@ -73,9 +73,9 @@ func DeleteSecretsForTenant(tenant string) error {
 func StoreAccessToken(tenant, value string) error {
 	// First, clear any existing chunks to prevent concatenation issues.
 	for i := 0; i < secretAccessTokenMaxChunks; i++ {
-		if err := keyring.Delete(secretClientSecret, tenant); err != nil {
+		if err := keyring.Delete(fmt.Sprintf("%s %d", secretAccessToken, i), tenant); err != nil {
 			if !errors.Is(err, keyring.ErrNotFound) {
-				return fmt.Errorf("failed to delete client secret from keyring: %s", err)
+				return fmt.Errorf("failed to delete access token chunk from keyring: %s", err)
 			}
 		}
 	}
--- a/internal/keyring/keyring_test.go
+++ b/internal/keyring/keyring_test.go
@@ -87,6 +87,42 @@ func TestSecrets(t *testing.T) {
 		assert.Equal(t, expectedAccessToken, actualAccessToken)
 	})
 
+	t.Run("it overwrites a previously stored token without leaving stale chunks", func(t *testing.T) {
+		keyring.MockInit()
+
+		// Store a long token spanning several chunks.
+		longToken := randomStringOfLength((2048 * 5) + 1)
+		err := StoreAccessToken(testTenantName, longToken)
+		assert.NoError(t, err)
+
+		// Overwrite with a token that fits in a single chunk.
+		shortToken := "short-token"
+		err = StoreAccessToken(testTenantName, shortToken)
+		assert.NoError(t, err)
+
+		// Must return exactly the short token, with no trailing chunks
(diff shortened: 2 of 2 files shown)


ramya18101 (approved):

ramya18101:
Hi @sparkling 👋, This is a great catch!

Thanks for finding this and putting together.
Really appreciate you digging through the old PR and tracking down the root cause here.

Also please sign your commit!

ramya18101 (approved):

PR #1534, opened 2026-05-28
feat(event-streams): improve subscribe UX

developerkunal:
### 🔧 Changes

Improves the `auth0 event-streams subscribe` UX so live event inspection is easier to read, easier to interrupt, and easier to resume, and keeps long-lived sessions alive across the server's periodic connection rotations.

**Discoverability**
- `--list-event-types` prints every value accepted by `--event-type` and exits without opening a subscription.
- Invalid `--event-type` values are now collected up front and reported in a single error listing every bad value, instead of bailing on the first. The error points at `--list-event-types` for the full list, and the SDK's internal Go type name no longer leaks into user output.

**Reconnect behavior**
- The server rotates long-lived SSE connections every few minutes. Previously this exhausted the SDK's lifetime reconnect cap and the session silently ended. The command now owns reconnection via a CLI-level resume loop that re-subscribes from the last cursor, so the session stays continuous and no events are missed.
- Reconnect delay is exponential backoff with full jitter, accrued only on consecutive zero-progress failures; any forward progress resets it, so healthy rotations reconnect instantly and silently. The server-advertised SSE `retry:` directive is honored as a floor, and a 1s minimum interval between attempts prevents a flapping connection from spinning in a tight loop.
- `--no-reconnect` exits cleanly when the connection ends instead of resuming.
- `--max-reconnects N` caps consecutive failed reconnect attempts before giving up (0 = retry indefinitely while progress is being made). Mutually exclusive with `--no-reconnect`.
- Ctrl+C cancels via context so the disconnect summary always prints.

**Output cleanups**
- `connection_timeout` frames (a protocol artifact emitted right before the server drops the SSE connection) are filtered up front, so the rendered output, `--json` / `--json-compact`, and the `--output-file` capture all stay consistent and free of reconnect noise.
- Server-emitted `error` events (rate limits, auth, etc.) render on their own row, are tracked in a separate `Errors:` counter in the summary, and no longer pollute per-type event counts.
- The disconnect summary now reports session duration, an `Errors:` line when non-zero, and groups per-type counts under a `By type:` heading, sorted by count desc with aligned columns.
- Event rows align: type and source pad to fixed widths before ANSI coloring so escape codes don't break visual alignment. Event IDs and error messages stay full-length so nothing forensic is truncated.
- The resume hint prints as a single line so the cursor is easy to copy.

**Bumps**
- `github.com/auth0/go-auth0/v2`: v2.11.0 → v2.12.0.

### 📚 References

None.

### 🔬 Testing

Unit tests added in `internal/cli/event_streams_subscribe_test.go` covering the pure helpers:

- `summarizeEvent` for heartbeat, error, `connection_timeout`, and concrete-event envelope variants.
- `padRight` for short, equal, long, and empty inputs.
- `shortOffset` for short, boundary, and long cursor truncation.
- `colorForEventType` confirms the original type string survives the color wrap for every suffix branch plus `error` and unknown.
- `invalidEventTypesError` covers single-value and multi-value cases, plus a guard that the SDK's internal type name never leaks into user output.
- `reconnectBackoff` confirms an immediate reconnect on a healthy rotation, that jittered delays stay within their exponential ceiling and never exceed the 30s cap, and that the server `retry:` directive acts as a floor.

Run with:

```
go test ./internal/cli/ -run 'TestSummarizeEvent|TestPadRight|TestShortOffset|TestColorForEventType|TestInvalidEventTypesError|TestReconnectBackoff'
```

Manual end-to-end testing against a live tenant:

- `auth0 event-streams subscribe` left running across multiple server connection rotations to confirm the session stays continuous with no event loss and no reconnect noise, then Ctrl+C to confirm the summary shows duration, event counts, errors when any, heartbeats, the `By type:` block, and a readable resume hint.
- `auth0 event-streams subscribe --list-event-types` to confirm the full list prints and the command exits.
- `auth0 event-streams subscribe --event-type bad.value --event-type also.bad` to confirm both bad values are reported in one error.
- `auth0 event-streams subscribe --no-reconnect` and `--max-reconnects 1` to confirm both flags are accepted and mutually exclusive.

### 📝 Checklist

### DIFF
--- a/internal/cli/event_streams_subscribe.go
+++ b/internal/cli/event_streams_subscribe.go
@@ -388,7 +388,9 @@ func subscribeEventStreamCmd(cli *cli) *cobra.Command {
 					return nil
 				}
 
+				attemptStart := time.Now()
 				progressed, serverRetry, err := runStreamSession(ctx, cli.apiv2, req, lastOffset, handleEvent)
+				sessionDuration := time.Since(attemptStart)
 
 				// Ctrl+C or parent cancellation: always a graceful exit.
 				if userInterrupted.Load() || errors.Is(err, context.Canceled) || ctx.Err() != nil {
@@ -423,6 +425,14 @@ func subscribeEventStreamCmd(cli *cli) *cobra.Command {
 				}
 
 				delay := reconnectBackoff(consecutiveFailures, serverRetry)
+				// Guard against a flapping connection that delivers a frame and
+				// dies immediately: each such session "progresses" and resets
+				// the backoff, which would otherwise spin in a tight reconnect
+				// loop. Enforce a minimum spacing between attempts based on how
+				// long the session actually lasted, regardless of progress.
+				if remaining := reconnectMinInterval - sessionDuration; remaining > delay {
+					delay = remaining
+				}
 				// Healthy connection rotations resume silently (even when the
 				// server asks for a short `retry:` wait) so a normal long-lived
 				// session looks seamless. Only surface a notice once we're
@@ -533,6 +543,10 @@ func reconnectBackoff(consecutiveFailures int, serverRetry time.Duration) time.D
 const (
 	reconnectBaseBackoff = 500 * time.Millisecond
 	reconnectMaxBackoff  = 30 * time.Second
+	// Floor between consecutive reconnect attempts regardless of progress, so a
+	// connection that flaps (delivers a frame then dies immediately) can't spin
+	// in a tight reconnect loop.
+	reconnectMinInterval = 1 * time.Second
 )
 
 // eventSummary is a generic, payload-agnostic projection of an SSE message
--- a/internal/cli/event_streams_subscribe.go
+++ b/internal/cli/event_streams_subscribe.go
@@ -1,10 +1,12 @@
 package cli
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


internal/cli/event_streams_subscribe.go
```diff
@@ -167,23 +246,50 @@ func subscribeEventStreamCmd(cli *cli) *cobra.Command {
...
 					if lastOffset != "" {
+						// The cursor is opaque and often very long; print the
+						// command and cursor on separate lines so the cursor
+						// wraps cleanly and is easy to select / copy in a
+						// terminal.
 						cli.renderer.Newline()
-						cli.renderer.Infof("Resume with: %s", ansi.Cyan(fmt.Sprintf("auth0 event-streams subscribe --from %s", lastOffset)))
+						cli.renderer.Infof("Resume from cursor %s:", ansi.Faint(shortOffset(lastOffset)))
+						cli.renderer.Output("  " + ansi.Cyan("auth0 event-streams subscribe --from \\"))
+						cli.renderer.Output("    " + ansi.Cyan(lastOffset))
```
ramya18101: We are adding lastOffSet twice in the output.. one in the info and another in the output command.. Can this be reverted..?

duedares-rvj (dismissed):

duedares-rvj (approved):

PR #1585, opened 2026-07-10
feat: add `commands` to list the command tree for easier discovery

developerkunal:
### 🔧 Changes

Makes the CLI easy to discover and drive programmatically, both for people and AI agents.

**`auth0 commands`** — a new top-level command that prints the entire CLI command tree with a short description of each command, so the whole surface is discoverable in one place instead of inspecting each `--help` page. Output shapes:
- **Default** nested tree, easy to scan visually.
- **`--flat`** every runnable command on its own line, easiest to scan or match an intent against.
- **`--json`** machine-readable; with **`--detailed`** it also includes the usage line, positional arguments, local flags (name, shorthand, usage, type, default), aliases, and whether authentication is required, enough to construct a valid invocation programmatically.
- Accepts an optional command path (for example `auth0 commands apps create`) to scope to one branch, and `--depth` to limit levels. Requires no authentication.

**JSON help for any command** — combining `--help` with `--json` prints that command's help as JSON (the same detailed shape as above, plus a `note` that steers an agent to `auth0 api` when a needed field is not exposed as a flag). This is intercepted before Cobra parses flags, so it works for every command, including the root and namespace commands that have no `--json` flag of their own.

**Agent mode** — setting `AUTH0_AGENT_MODE=1` (any truthy value) makes `--help` emit JSON without needing `--json`, which is convenient for scripts and AI agents.

**Discoverability touches** — the usage footer now points to `auth0 commands` and JSON help, and the homepage docs gain a "Discovering and Scripting Commands" section.

No new dependencies, and the tree-walking runs only when these paths are invoked, so there is no impact on the startup or runtime of other commands.

### Demo 

pr-demo

### 📚 References

N/A

### 🔬 Testing

- Unit tests in `internal/cli/commands_test.go` cover tree building, depth limiting, detailed output, flat output (including the scoped case), positional-argument extraction that ignores flag values, help/JSON request detection, agent-mode env parsing, the fallback-note annotation, and the JSON-help interception (explicit `--help --json`, agent mode without `--json`, and the compact root overview).
- The full `internal/cli` suite passes and `go vet` is clean.
- Regenerated the docs so `docs/auth0_commands.md` and `docs/index.md` stay in sync.

Manual checks:

```
auth0 commands
auth0 commands --flat
auth0 commands apps show --json --detailed
auth0 apps create --help --json
AUTH0_AGENT_MODE=1 auth0 apps create --help
```

### 📝 Checklist


ramya18101:
Thanks for raising the PR!

Before we route agents to the subcommands, we should ensure that they provide the same functionality as `auth0 api` does today, specifically, the ability to accept the complete request schema as a payload.

ramya18101 (approved):

PR #1587, opened 2026-07-14
feat: add session-transfer delegation flags in apps (EA only)

bkiran6398:
### 🔧 Changes

Adds support for configuring the Early Access **session-transfer delegation** settings on an application, used for Custom Token Exchange (Impersonation via Session Transfer). This surfaces the new `client.session_transfer.delegation` Management API object in the CLI.

Two new flags on `auth0 apps session-transfer update`:

- `--delegation-allow-delegated-access` (bool) — allow the application to accept Session Transfer Tokens containing an Actor, enabling delegated (impersonation) access. Defaults to `false`.
- `--delegation-enforce-device-binding` (string) — device binding enforcement for delegated access. Accepts `ip` or `asn` only (defaults to `ip`).

Usage:

```
auth0 apps session-transfer update <app-id> --delegation-allow-delegated-access=true --delegation-enforce-device-binding=asn
```

Behavior notes:

- Delegation is sent to the API only when one of its flags is explicitly set, so updates to the existing session-transfer settings leave delegation untouched. Clients not using the feature are unaffected.
- The two delegation fields can be updated independently; the API merges within the delegation object, so setting one leaves the other intact.
- `--delegation-enforce-device-binding` is validated as `ip` or `asn` before any API call. Unlike the parent `--enforce-device-binding`, it does not accept `none`, matching the Management API enum.
- `apps session-transfer show`/`update` output includes the delegation fields only when the client has delegation configured, keeping output unchanged for existing clients.

The two fields are Early Access and gated by the `cte_session_transfer_token` tenant feature flag on the API side.

### 📚 References

- Related terraform-provider-auth0 PR: https://github.com/auth0/terraform-provider-auth0/pull/1618
- SDK support (go-auth0 v1): https://github.com/auth0/go-auth0/pull/803

### 🔬 Testing

- **Unit tests:** Added display tests covering the delegation rows rendering when configured and their omission when absent.
- **Integration tests:** Added a case asserting `--delegation-enforce-device-binding` rejects an invalid value (`none`). The successful update/show happy path is covered by unit tests rather than a live-tenant integration case, because delegation is gated behind the `cte_session_transfer_token` Early Access flag, which is not guaranteed on the CI test tenant.
- **Manual verification (live tenant with the EA flag enabled):**
  - Set both fields, then updated only one and confirmed the other was preserved (independent updates, both directions).
  - Confirmed updating a legacy session-transfer flag does not introduce a delegation object for a client without one.
  - Confirmed `session-transfer show` renders the delegation rows only when configured.
  - Confirmed the API rejects delegation for clients with token-endpoint auth method `none`; the CLI surfaces this error verbatim.

To test end-to-end (requires a tenant with `cte_session_transfer_token` enabled and a confidential client):

```
auth0 apps session-transfer update <app-id> --delegation-allow-delegated-access=true --delegation-enforce-device-binding=asn --json
auth0 apps session-transfer update <app-id> --delegation-enforce-device-binding=ip --json   # allow_delegated_access preserved
auth0 apps session-transfer show <app-id>
```

### 📝 Checklist


internal/cli/apps.go
```diff
@@ -164,6 +164,18 @@ var (
 		Help:         "Device binding enforcement: 'none', 'ip', or 'asn'.",
 		AlwaysPrompt: true,
 	}
+	appSTDelegationAllowAccess = Flag{
+		Name:     "Allow Delegated Access",
+		LongForm: "delegation-allow-delegated-access",
```
ramya18101: See if we can add shorter aliases for both of these flags, since we couldn't use the standard short-form flags.

internal/cli/apps.go
```diff
@@ -1235,6 +1250,16 @@ func appsSessionTransferUpdateCmd(cli *cli) *cobra.Command {
 				inputs.ID = args[0]
 			}
 
+			// Validate the delegation device binding before any API call.
+			if appSTDelegationDeviceBinding.IsSet(cmd) &&
```
ramya18101: We don't add validations on cli v1, Let the API return the error.

internal/cli/apps.go
```diff
@@ -1283,6 +1308,22 @@ func appsSessionTransferUpdateCmd(cli *cli) *cobra.Command {
 				st.EnforceDeviceBinding = current.SessionTransfer.EnforceDeviceBinding
 			}
 
+			// Delegation (EA) is sent only when a flag is set, leaving it untouched for
+			// others. The API merges sub-fields, so sending just the changed one is enough.
+			if appSTDelegationAllowAccess.IsSet(cmd) || appSTDelegationDeviceBinding.IsSet(cmd) {
```
ramya18101, line 1307: Where is the AskU logic..?

ramya18101 (changes requested):

internal/cli/apps.go
ramya18101: @bkiran6398, Is there a reason why we don't prompt for the newly added flags  in the `auth0 apps session-transfer update <app-id>` flow?

bkiran6398: Delegation params are EA-gated, so prompting every user for it would expose a setting that most can't use and lead to a `400` error response if provided.

ramya18101: Got it, Please highlight this [example](https://github.com/auth0/auth0-cli/pull/1587/changes#diff-3bc90f8ea54bc289631b44849d75d342573625452a6fde0144e7e2d98be80725R1243) as EA

ramya18101 (approved):

PR #1607, opened 2026-08-10
fix(up): rename portal pages property from content to standard

jacobovidal:
## Summary

  Renames the `portalPages.Content` JSON tag from `"content"` to `"standard"` to match the portals API.

 Related PR: https://github.com/auth0/auth0-cli/pull/1595

  ## Test plan

  - [ ] Run `auth0 universal-portals setup` end-to-end and confirm portal is created successfully.


arpit-jn (approved):
LGTM

PR #1665, opened 2026-09-16
feat: emit fully machine-readable JSON output in agent mode

developerkunal:
### 🔧 Changes

Builds on #1663 to make the CLI's output fully machine-readable when it runs in agent mode (auto-detected, or forced with `--agent-mode` / `AUTH0_AGENT_MODE`), and whenever `--json` is used where it applies. This is the JSON-output layer on top of the exit-code and error-envelope work in the base PR.

- **Streaming as JSON (`internal/display`).** Streaming commands now emit a sequence of newline-separated JSON objects instead of a JSON array, so a reader can consume records incrementally without waiting for an array that never closes while tailing. Agent mode and `--json-compact` emit compact newline-delimited JSON (one object per line); `--json` indents each object for a human watching the live stream. `auth0 logs tail` is the one streaming command and previously rendered a table.
- **`--json` / `--json-compact` on `auth0 logs tail`.** These flags were never registered on `tail`, so `auth0 logs tail --json` failed with `unknown flag`. Both are now supported and mutually exclusive.
- **Trailing newline on JSON stdout.** JSON results on stdout always end with a trailing newline, so a piped or newline-delimited reader never drops the final record.
- **Machine-clean stderr in agent mode.** In agent mode the CLI suppresses human diagnostics (info, success, detail, warning, and non-fatal errors) entirely, along with the decorative `=== <tenant> …` heading and blank spacing lines. On success stderr is empty and on failure it carries only the JSON error envelope from #1663. Because the envelope is the only thing on stderr, a merged stdout+stderr stream stays parseable, discriminated by the reserved `error` key rather than a fragile per-message field. The decorative heading and blank lines are also suppressed in any JSON output mode (`--json` and `--json-compact`). Normal (non-agent, non-JSON) output is unchanged, so a human running `--json` still sees hints on stderr.
- **Help as JSON in agent mode.** A bare `auth0` and a command group invoked without a subcommand (for example `auth0 apps`) return the JSON command tree rather than the human help text, and the root help now describes agent mode itself (its output contract and how to disable it) so it does not need to be re-announced on every command.

Documentation is updated in `CHANGELOG.md` and the regenerated `docs/`.

### 📚 References

Builds on #1663 (`feat/agent-error-contract`), which landed the exit-code and error-envelope contract this extends.

### 🔬 Testing

- Table-driven unit tests added for NDJSON vs indented streaming, the agent-mode compact contract, agent-mode stderr suppression (diagnostics and the blank spacing line), heading suppression, the trailing newline on JSON stdout, and JSON help for bare and namespace invocations.
- `go build ./...`, `go vet ./internal/...`, and the full unit-test suite pass; `make docs` is in sync.
- Verified live end-to-end against a real tenant: `auth0 apps create --agent-mode` prints the result object on stdout with an empty stderr, and `auth0 apps <unknown-subcommand> --agent-mode` prints nothing on stdout and a single `{"error":{…}}` line on stderr.
- Not run: the integration suite hits a live tenant. Existing `exactly: "[]"` assertions are unaffected because the runner trims a trailing newline before matching. `auth0 logs tail` itself was exercised manually rather than in CI, since it streams from a live tenant.

### 📝 Checklist


internal/cli/root.go
```diff
@@ -43,6 +43,26 @@ The Auth0 CLI now includes features for AI agents and automation:
...
+AUTH0_AGENT_MODE=false).
+
+In agent mode the CLI:
+
+  • Prints results to stdout as JSON, and streams (for example 'auth0 logs tail')
+    as newline-delimited JSON, one object per line.
+  • Prints diagnostics to stderr as JSON lines ({"level","message"}) and errors as
+    a JSON envelope ({"error":{"code","message","status","details"}}).
+  • Disables interactive prompts and colors.
+  • Exits with a code per failure class: 0 success, 1 generic, 2 usage, 3 auth,
```
ramya18101: Update this

ramya18101 (approved):

PR #1676, opened 2026-09-17
refactor: route agent/JSON command help through Cobra's help func

developerkunal:
### 🔧 Changes

Replaces the hand-rolled JSON-help pre-parser with Cobra's `SetHelpFunc`. In agent/JSON mode, `--help`, the `help` subcommand, and a bare namespace (whose `RunE` calls `Help()`) now emit the machine-readable JSON command tree, with the target command already resolved and its flags parsed by Cobra. Human help is unchanged.

Adds a narrow `renderNamespaceJSONHelp` guard for the one case Cobra's help func cannot reach: an explicit `--json`/`--json-compact` on a namespace, which defines no such flag and would otherwise be rejected as an unknown flag before help runs. The guard delegates flag reasoning to Cobra's own parser rather than re-implementing pflag, and falls through to Cobra for leaf commands, mistyped subcommands, and unknown flags so they run or error exactly as before.

Resolves agent mode up front in `Execute` so JSON help works on the `--help` path (which skips `PersistentPreRunE`) and the JSON error envelope still covers flag-parse failures. Net removal of ~90 lines of hand-rolled parsing.

### 🔬 Testing

- `make test-unit` (the `internal/cli` suite passes, including the new `TestRenderNamespaceJSONHelp`, `TestRenderCommandHelpJSON`, and end-to-end `TestJSONHelpFunc`).
- Verified against the built binary: namespace `--json`/`--json-compact` and root `--json` render JSON help (exit 0); `--help`, bare namespace, and the `help` subcommand render JSON help in agent mode and human help otherwise; `--version` still prints the version; unknown flags and mistyped subcommands still return the usage error envelope (exit 1); a leaf `--json` still runs the command.

### 📝 Checklist


ramya18101:
LGTM

